PR #2, opened 2024-05-30
MGMT-17783 Handle errors for labels with missing keys

celdrake:
Adds validation to labels to ensure the have the mandatory `key` property.
In case there are invalid labels, the message will indicate all the labels (values) which are missing the keys.

label-keys-validation

### DIFF
--- a/libs/ui-components/src/components/form/validations.ts
+++ b/libs/ui-components/src/components/form/validations.ts
@@ -1,10 +1,8 @@
 import * as Yup from 'yup';
 import { TFunction } from 'i18next';
+import { FlightCtlLabel } from '../../types/extraTypes';
 
-export interface UnvalidatedLabel {
-  key?: string;
-  value?: string;
-}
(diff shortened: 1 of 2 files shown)


libs/ui-components/src/components/form/validations.ts
```diff
@@ -1,18 +1,33 @@
 import * as Yup from 'yup';
 import { TFunction } from 'i18next';
 
-import { FlightCtlLabel } from '../../types/extraTypes';
+export interface UnvalidatedLabel {
```
rawagner: Lets use `Partial<FlightCtlLabel>` instead of creating a new type

rawagner (changes requested):

libs/ui-components/src/components/form/validations.ts
celdrake: Oh, yes. 

I still defined the type in this file to avoid having to define it in each usage, is that OK?

rawagner (approved):

PR #6, opened 2024-06-05
MGMT-17987 Edit device labels inline

celdrake:
Editing labels of a device is done in the Device Details page, and the "Edit labels" modal is deleted.

I added a longish time (2s) for debouncing the request, so if labels are quickly changed, a single request is performed. 
While there may be validation errors, no API calls are done.

~~Had to set the base `patternfly/react-core` package to version `^5.3.x` because this is where the possibility to disable the "add label" button was first introduced.~~ Kept existing Patternfly version based on Code Review feedback

It's used to disable the "add label" button while a submit is in progress.
[edit-labels-inline.webm](https://github.com/flightctl/flightctl-ui/assets/829045/269b70d4-af73-41fd-8dfd-1dc990fcffdc)


libs/ui-components/package.json
```diff
@@ -27,7 +27,7 @@
     "yup": "^1.3.3"
   },
   "peerDependencies": {
-    "@patternfly/react-core": "^5.0.0",
+    "@patternfly/react-core": "^5.3.0",
```
rawagner: we should not go higher than OCP react-core https://github.com/openshift/console/blob/master/frontend/package.json#L156 otherwise in our ACM plugin, new features from react-core wont be available

celdrake: Right, we should keep a consistent experience, plus with a greater version we could introduce bugs without noticing.

I've changed some settings on the label to simulate the behaviour from the new `isDisabled` prop.

PR #9, opened 2024-06-13
Updated device details page

celdrake:
Updates the device details page to the latest design, though some sections are not available yet and have been omitted for now.

Layout and content changes:

1. Basic details: OK
2. System status:
- "Application status" and "Update status": Omitted, not in the API.
- "Device status": Shown, though the statuses are not yet those from the DDD
- "Last seen": Not available yet, we're still showing "Created at" (to be changed soon)
3. Resource status:
Omitted, not in the API.
4. Configurations:
- "System image": OK
- "Fleet name": OK
- "Source name": Moved after fleet name, as there can be multiple. Renamed to "Sources" and kept the same component as in the Fleet details page.
- "Device overrides": Omitted, needs clarification
5. Applications --> renamed from "Containers", most information not available on the API yet
6. System services --> renamed from "Systemd units". It will show a table with the selected patterns, and a status (either Unknown or Ready). The status will need adjustments when the API statuses are available.

Additionally, the labels have been changed from "blue" to "no color" according to the new designs.
device-details-new-2


libs/ui-components/src/components/DetailsPage/DetailsPageCard.tsx
```diff
@@ -1,10 +1,17 @@
...
 import './DetailsPageCard.css';
 
 const DetailsPageCard: React.FC<CardProps> = (props) => <Card className="fctl-details-page__card" {...props} />;
 
+export const DetailsPageCardTitle = ({ title }: { title: string }) => (
+  <CardTitle className="fctl-details-page__card-title">
+    {title}
+    <Divider component="div" />
+  </CardTitle>
+);
```
rawagner: instead of adding a custom css, we could use `Stack`
```suggestion
  <CardTitle>
    <Stack hasGutter>
      <StackItem>
        {title}
      </StackItem>
      <StackItem>
        <Divider component="div" />
      </StackItem>
    </Stack>
  </CardTitle>
);
```

celdrake: I added the custom CSS because the divider line width seemed to not work well without that, but now I realize that it depends a bit on the viewport width, so it's not really needed.

But for reusability, wouldn't it be preferable to have our own component? The Overview page seems to have a very similar pattern, so I'd rather not DRY (actually, I'd argue that the pattern in Overview and Device details should be the same).

Is is OK if I remove the custom CSS and keep the component?

celdrake: I removed the custom component with the divider and its CSS based on the discussion on Slack.

libs/ui-components/src/components/DetailsPage/DetailsPageCard.css
```diff
@@ -6,3 +6,7 @@
 .fctl-details-page__card {
     height: 100%
 }
+
+.fctl-details-page__card-title .pf-v5-c-divider {
```
rawagner: the css rule can be removed too

rawagner (changes requested):

rawagner (approved):

PR #10, opened 2024-06-14
Adjust label selector display in FleetList

celdrake:
In FleetList, if a fleet has many labels, or they are long, the column for the label grew accordingly.
To prevent that, we add a new variant to the  `LabelsView` which sets a maximum width on larger viewports.

labels-2
labels-1

**Note**: This PR also fixes an issue whereby most of the forms including a LabelsField would not allow to edit or delete the labels after adding them.

### DIFF
--- a/libs/ui-components/src/components/common/LabelsView.tsx
+++ b/libs/ui-components/src/components/common/LabelsView.tsx
@@ -24,9 +24,7 @@ const LabelsView = ({ variant, prefix, labels }: LabelsViewProps) => {
       numLabels={isLimited ? 2 : 5}
       className={isLimited ? 'fctl-labels-view__collapsed' : undefined}
       expandedText={t('Show less')}
-      // Label group will use this template and inject the correct count of labels.
-      // For it to work, the variable name must be "remaining"
-      collapsedText={t('{{remaining}} more', { remaining: 0 })}
+      collapsedText={'${remaining} ' + t('more')}
     >
       {labelItems.map(([key, value], index: number) => (
         <Label key={`${prefix}_${index}`} id={`${prefix}_${index}`} color="blue">
--- a/libs/ui-components/src/components/common/LabelsView.tsx
+++ b/libs/ui-components/src/components/common/LabelsView.tsx
@@ -19,13 +19,14 @@ const LabelsView = ({ variant, prefix, labels }: LabelsViewProps) => {
   }
 
   const isLimited = variant === 'collapsed';
-
   return (
     <LabelGroup
       numLabels={isLimited ? 2 : 5}
       className={isLimited ? 'fctl-labels-view__collapsed' : undefined}
       expandedText={t('Show less')}
-      collapsedText={t('${remaining} more')}
+      // Label group will use this template and inject the correct count of labels.
+      // For it to work, the variable name must be "remaining"
+      collapsedText={t('{{remaining}} more', { remaining: 0 })}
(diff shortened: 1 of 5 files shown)


libs/ui-components/src/components/common/LabelsView.tsx
```diff
@@ -1,14 +1,32 @@
...
+
+  const isLimited = variant === 'collapsed';
+
   return (
-    <LabelGroup numLabels={5}>
+    <LabelGroup
+      numLabels={isLimited ? 2 : 5}
+      className={isLimited ? 'fctl-labels-view__collapsed' : undefined}
+      expandedText={t('Show less')}
+      collapsedText={t('${remaining} more')}
```
rawagner: `remaining` needs to be passed as property `count`

libs/ui-components/src/components/form/LabelsField.tsx
```diff
@@ -15,10 +15,16 @@ type LabelsFieldProps = {
   onChangeCallback?: (newLabels: FlightCtlLabel[], hasErrors: boolean) => void;
 };
 
-const LabelsField: React.FC<LabelsFieldProps> = ({ name, onChangeCallback, addButtonText, isEditable }) => {
+const LabelsField: React.FC<LabelsFieldProps> = ({
+  name,
+  onChangeCallback,
+  addButtonText,
+  isEditable: isEditableProp,
```
rawagner:
```suggestion
  isEditable = true,
```

then we dont need the other changes below I guess

rawagner (changes requested):

libs/ui-components/src/components/common/LabelsView.tsx
celdrake: Actually it's not needed because the `LabelGroup` uses this text as a template and adds the number itself.
However, this could cause confusion to the translators, so I will change it as if it was a normal label so it works as expected.

libs/ui-components/src/components/form/LabelsField.tsx
celdrake: True, changed!

libs/ui-components/src/components/common/LabelsView.tsx
```diff
@@ -1,14 +1,33 @@
...
+  const isLimited = variant === 'collapsed';
   return (
-    <LabelGroup numLabels={5}>
+    <LabelGroup
+      numLabels={isLimited ? 2 : 5}
+      className={isLimited ? 'fctl-labels-view__collapsed' : undefined}
+      expandedText={t('Show less')}
+      // Label group will use this template and inject the correct count of labels.
+      // For it to work, the variable name must be "remaining"
+      collapsedText={t('{{remaining}} more', { remaining: 0 })}
```
rawagner: hmm, is this working now ? the `t` function will return a string `0 remaining` so I dont think PF will replace the 0.

I think it should be something like 
```suggestion
      collapsedText={'${remaining} '+t('more')}
```

`collapsedText` prop is is very i18n unfriendly :/

celdrake: Yes, it's a tricky one. It worked well when passing the pattern with `${remaining}`, but now I see it doesn't work with the proper `i18n` label.

I like your suggestion better to avoid confusing translators and make PF happy as well.

rawagner (approved):

PR #11, opened 2024-06-17
MGMT-18181 Add basic restrictions to fields

celdrake:
This PR adds the basic UI validations that will be required when https://github.com/flightctl/flightctl/pull/289/files is merged.

Validations added in the first phase:
- Adds restrictions to fields which should be valid k8s labels (repository name, configuration names, etc)
- Fixes the request for a new fleet when it doesn't specify an OS image.
- Adds info for the inline configs, that the YAML content must be a valid ignition file. In the future, we'll need to provide better validation for this field (maybe by having a dedicated endpoint for validating it, if not possible to do it via the UI itself).
validations

Validations added in the second phase:
- max-length restrictions
- systemd units, make sure they match the expected regular expression
- list of labels (device selector labels, fleet labels, etc): make sure they are valid k8s labels
multiple-labels

I think this covers all the required validations from the API's PR, and in theory the PR in the UI could be merged prior to the API's PR. (We'll be more restrictive than allowed at the moment).


celdrake:
@rawagner the Backend PR has now 2 approvals, please review this one when you have time.

libs/cypress/fixtures/enrollmentRequests/index.ts
```diff
@@ -5,7 +5,7 @@ const approvedErStatus = {
     approved: true,
     approvedAt: '2024-04-10T12:13:45Z',
     approvedBy: 'unknown',
-    labels: { displayName: 'east-device', fleet: 'eu-east-prod-001', region: 'spain' },
+    labels: { 'display-name': 'east-device', fleet: 'eu-east-prod-001', region: 'spain' },
```
rawagner: we dont need to change this. `displayName` is valid label key

libs/ui-components/src/components/Repository/CreateRepository/utils.ts
```diff
@@ -278,23 +282,25 @@ export const getRepositoryPatches = (values: RepositoryFormValues, repository: R
...
-          .defined(t('Name is required.'))
-          .test('Must be unique', t('Must be unique'), (value) => {
-            if (!value) {
-              return true;
-            }
-            return values.filter((v) => v.name === value).length === 1;
-          }),
-        targetRevision: Yup.string().defined(t('Target revision is required.')),
-        path: Yup.string().matches(pathRegex, t('Must be an absolute path.')).defined(t('Path is required.')),
+        name: validKubernetesLabel(t, { isRequired: true }).test('Must be unique', t('Must be unique'), (value) => {
```
rawagner, line 291: we should have a name-specific validation. The valid values are a bit different - 
 we should link https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names (dns subdomain is also used by fctl backend)

libs/ui-components/src/components/form/validations.ts
```diff
@@ -4,7 +4,26 @@ import { FlightCtlLabel } from '../../types/extraTypes';
 
 type UnvalidatedLabel = Partial<FlightCtlLabel>;
 
-export const uniqueLabelKeysSchema = (t: TFunction) =>
+const SYSTEMD_PATTERNS_REGEXP = /^[a-z][a-z0-9-_.]*$/;
+const SYSTEMD_UNITS_MAX_PATTERNS = 256;
+
+const K8S_LABEL_REGEXP = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/;
```
rawagner, line 10: This regex is incorrect IMO.

we need to split it as different parts of label have different requirements
key - `([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]`
key prefix - `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`
value - `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`

also lengths of each part needs to be verified as described in https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

rawagner (changes requested):

libs/cypress/fixtures/enrollmentRequests/index.ts
celdrake: You are right, I'll revert these changes.

libs/ui-components/src/components/Repository/CreateRepository/utils.ts
celdrake: I see. I suggest to continue with those validations in the follow-up for the interactive checker.

With this PR, we'd make sure that the UI won't have 400 errors, and I'll continue handling the the more nuanced validations for labels there.

libs/ui-components/src/components/form/validations.ts
celdrake: Agree. As before, since this requires additional checks and UX, I can handle that separately.

rawagner (approved):

libs/ui-components/src/components/Repository/CreateRepository/utils.ts
rawagner: Ok, lets improve in a follow-up

PR #23, opened 2024-07-03
DeviceEnrollment: Remove region field, make displayName optional

celdrake:
- The "region" field for approving an Enrollment Request is not mandatory in the API anymore, however it still must be sent in the request or it fails.

In preparation for the field being dropped, we can remove it from the form and send an empty region.

- The displayName is now optional


rawagner:
Waiting for https://github.com/flightctl/flightctl/pull/331

rawagner:
backend patch is merged. We should be able to remove the region field from the request too.

celdrake:
> backend patch is merged. We should be able to remove the region field from the request too.

Great, I also entirely removed it from the UI now.

rawagner (approved):

PR #25, opened 2024-07-04
MGMT-18323 unify repository page without tabs

celdrake:
It was discussed that the Repository page should not use tabs, and instead all content should be in the main page.

Adapted the URL to link to a particular resource sync, now the URL will be in the form devicemanagement/repositories/default-repo#<resourcesync-name>

repo-page


rawagner (approved):

rawagner (changes requested):
We need to remove Resource Syncs tab (navigation) from Repository Details page

celdrake:
> We need to remove Resource Syncs tab (navigation) from Repository Details page

Fixed, I left them by mistake.

rawagner (approved):

PR #28, opened 2024-07-05
MGMT-17948 Show number of devices selected by a new fleet

celdrake:
In the Create new fleet wizard, now the "device selector" will show the number of selected devices based on the label match.

https://github.com/flightctl/flightctl-ui/assets/829045/08d1dad8-34e6-4ad9-ad02-3ac5b4e52eed


nirfarkas:
WOW! This looks great! Maybe we can make the number of selected devices more prominent? Let's brainstorm about this on Monday.

celdrake:
> WOW! This looks great! Maybe we can make the number of selected devices more prominent? Let's brainstorm about this on Monday.

Thanks! Changed the message according to the new designs, to show an alert when there are added labels.

libs/ui-components/src/components/Fleet/CreateFleet/steps/DeviceLabelSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <Alert isInline variant="danger" title={t('Failed to determine the number of selected devices')}>
+        {deviceCountError}
+        <Button
+          variant="link"
+          icon={<RedoIcon />}
+          onClick={() => {
+            reloadDeviceSelection(labels);
+          }}
+        >
+          Try again
```
rawagner: missing i18n

celdrake: Added

rawagner (approved):

PR #34, opened 2024-07-10
Use backend filters on Devices list and Overview pages

rawagner:


libs/ui-components/src/components/Device/DeviceList/DeviceTableToolbar.tsx
```diff
@@ -0,0 +1,127 @@
...
+      {Object.keys(activeStatuses)
+        .filter((k) => !!activeStatuses[k as keyof FilterStatusMap].length)
+        .map((k) => {
+          const key = k as keyof FilterStatusMap;
+          const { title, items } = getStatusItem(t, key);
+          return (
+            <SplitItem key={key}>
+              <ChipGroup categoryName={title} isClosable onClick={() => updateStatus(key)}>
+                {activeStatuses[key].map((status: string) => (
+                  <Chip key={status} onClick={() => updateStatus(key, status)}>
```
celdrake, line 101: Similar to what you did for the fleet name, I think we should also debounce the statuses.
Because at the moment, whenever a status checkbox is clicked, a new request is triggered immediately.

Since it's easy to click multiple statuses in short period of time, we could be triggering unnecessary requests.

We can add it as a follow-up task though.

libs/ui-components/src/components/Device/DeviceList/DeviceTableToolbar.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <DeviceToolbarChips {...rest} updateStatus={updateStatus} />
+    </>
+  );
+};
+
+type DeviceToolbarChipsProps = Omit<DeviceTableToolbarProps, 'setActiveStatuses'> & {
+  updateStatus: UpdateStatus;
+};
+
+const DeviceToolbarChips = ({
```
celdrake, line 81: We used to have a "Clear all filters" button.
Should we keep it?

celdrake (approved):

rawagner:
Thanks for the quick review @celdrake . I will create a followups for your suggestions

PR #62, opened 2024-07-25
EDM-252: Add debug information for device status

celdrake:
Allows the user to see the entire `device.status` information in order to debug issues.
Should be unnecessary when the status events are reported.

https://github.com/user-attachments/assets/d9d11e1b-4ec0-42f6-806b-9260bb262d69


hexfusion (commented):
love!

PR #66, opened 2024-07-29
EDM-257: Fix missing tooltip in device charts

rawagner:


rawagner:
@celdrake Updated to use the count instead of %

celdrake (approved):

PR #81, opened 2024-08-01
Fix wrong match to fleets with no label selector

celdrake:
The match selector was identifying fleet with no labels as matching all devices, when in reality, they don't match any devices.


celdrake:
Merging since it's a bug which is being seen in the demo environment.

PR #15, opened 2024-06-24
Link Microshift managed cluster

rawagner:
This is not yet done on the backend but ATM we expect that managed cluster name will equal device name.

Screenshot from 2024-06-24 12-09-46

PR #86, opened 2024-08-07
EDM-292: Show MicroShift cluster link

celdrake:
Continuation of https://github.com/flightctl/flightctl-ui/pull/15

Mostly added error handling, although I will work on adding `ErrorBoundary` to the OCP pages in a separate PR.

When the device has an associated MicroShift cluster:
without-error

When the device has an associated MicroShift cluster, but the k8s API returns an error:
with-error

(details appear when hovering over the message)

### DIFF
--- a/apps/ocp-plugin/src/components/Devices/DeviceDetailsPage.tsx
+++ b/apps/ocp-plugin/src/components/Devices/DeviceDetailsPage.tsx
@@ -1,8 +1,66 @@
 import * as React from 'react';
 import DeviceDetails from '@flightctl/ui-components/src/components/Device/DeviceDetails/DeviceDetailsPage';
+import { useK8sWatchResource } from '@openshift-console/dynamic-plugin-sdk';
+import { useAppContext } from '@flightctl/ui-components/src/hooks/useAppContext';
+import {
+  Alert,
+  Button,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  Spinner,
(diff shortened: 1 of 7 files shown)


apps/ocp-plugin/src/components/Devices/DeviceDetailsPage.tsx
```diff
@@ -1,8 +1,70 @@
...
+      kind: 'ManagedCluster',
+      group: 'cluster.open-cluster-management.io',
+      version: 'v1',
+    },
+    name: clusterName,
+  });
+
+  const watchResultError = getWatchK8sResourceResult(k8sError as K8sWatchResourceError, true);
+  if (k8sError && !watchResultError) {
+    return <DeviceDetails />;
```
rawagner, line 32: This is a bit weird. When we go to the details page we will see `Microshift cluster` item which is in loading state. And if it is not found, the whole item disappears. IMO we should have at at all times, and if we get 404, just show `-`

celdrake: Nir had mentioned that we shouldn't show the column when there's no MicroShift cluster.
So probably we shouldn't show the loader either, WDYT?

celdrake: Based on further discussion with Nir, we won't use the Spinner. 
And when there isn't an associated MicroShift cluster, we won't show the column either.

rawagner (approved):

PR #91, opened 2024-08-09
NO-ISSUE: Add dev:auth script and allow disabling TLS verification fo…

rawagner:
…r proxy


celdrake (approved):
Both authenticated and unauthenticated options worked perfectly.

PR #96, opened 2024-08-22
NO-ISSUE: Make sure that we push container and helm releases for tags

mangelajo:
Otherwise tagged versions won't generate the artifacts.


rawagner (approved):

mangelajo:
Thanks @rawagner that was quick!!!

PR #99, opened 2024-08-27
EDM-383 Show repository details if only RS failed to load

celdrake:
When attempting to edit a repository and the request to fetch its associated resource syncs failed, the UI was not showing any of the repository details.

This is happening ATM when the repository name is longer than 63 characters, because the API query for filtering the RSs by repository fails (I opened a bug for it).

In any case, the UI should show the repository details, but not allow modifications as any existing RS could be impacted by the changes.

no-edit-repository


libs/ui-components/src/components/Repository/CreateRepository/CreateRepository.tsx
```diff
@@ -108,6 +107,14 @@ const CreateRepository = () => {
             {repositoryId ? t('Edit repository') : t('Create repository')}
           </Title>
         </StackItem>
+        {rsError && (
+          <Alert isInline variant="warning" title={t('Failed to retrieve the resource syncs')}>
+            {t(
+              'We could not verify if this repository contains resource syncs. Changes to resource syncs are not available at the moment for this reason.',
```
rawagner: Im not very happy with the msg. How about 
```suggestion
              'We could not verify if this repository contains resource syncs. Changes to resource syncs are disabled.',
```

Can we also add a `try again` action ?

celdrake: - I thought of saying "disabled", but since I hid the section completely, it felt that could be confusing.
Would it be better to instead showing the "Use RS" checkbox as disabled and changing the message?

- I will add the "try again" action.

celdrake: I spoke to Nir, and we discussed that it may be better to make the entire Repository read-only when we fail to fetch the RSs, because if they existed, they could be affected by a change in the repository URL, etc.

rawagner (approved):

PR #102, opened 2024-08-29
EDM-384: Use new query to retrieve the repository resource syncs

celdrake:
To obtain the RSs belonging to a repository, we were adding a label to the RS, and querying RSs based on this label.
The query would fail if the repository name is not a label (eg. it is longer than 63 characters).

Now we will use a dedicated query which is being added in https://github.com/flightctl/flightctl/pull/501
which allows us to query irrespective of the repository name.

new-rs-query


celdrake:
@rawagner the backend PR was merged and I fixed the conflict issue, could you please review?

rawagner (approved):

PR #112, opened 2024-09-17
EDM-296: Simplify helm charts and enable global helm values

rawagner:
Notable changes:
 - look at `global` values. This will be useful once the UI charts will be used as a sub-charts of the main FCTL chart
 - namespace is taken from `.Release.Namespace`
 - non-secret values are stored in `ConfigMap` instead of `Secret`
 - qcow2 and bootc vars are removed as the feature is simply not there/worth exposing

This is a first part of being able to include UI chart as sub-chart to main FCTL chart


apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/values.yaml
```diff
@@ -1,12 +1,17 @@
...
+  image: quay.io/flightctl/flightctl-ocp-ui
+  pullPolicy: Always
+  tag: ""
+## @param api.url URL of Flightctl API endpoint accessible by UI backend.
+## @param api.grpcUrl URL of Flightctl GRPC endpoint accessible by UI backend.
+## @param api.insecureSkipTlsVerify Set to true if Flightctl TLS certificate validation should be skipped.
+## @param api.caCert A custom CA cert for Flightctl TLS.
+api:
+  url: https://flightctl-api:3443/
+  grpcUrl: grpcs://flightctl-api-agent-grpc:7443/
```
celdrake: I tested it locally and the terminal worked fine, but we were previously using port 7444.
Would port 7443 work also in the deployed environment?

apps/standalone/deploy/helm/flightctl-ui/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: flightctl-ui
-description: A helm chart for flightctl-ui and keycloak
+description: A helm chart for flightctl UI
 type: application
 version: 0.0.1
```
celdrake, line 5: Should we maybe up the `version`? (Same for the rest of the OCP plugin chart)

apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/values.yaml
```diff
@@ -1,12 +1,17 @@
-flightctlUi:
-  namespace: flightctl
-  image: quay.io/flightctl/flightctl-ocp-ui:latest
-  flightctlServer: https://flightctl-api:3443
-  flightctlMetricsServer: https://flightctl-api:9090
-  flightctlGrpcServer: flightctl-api-agent-grpc:7444
-  bootcImgUrl: "" #Bootc device image. Example: quay.io/example/example-agent-centos:bootstrap
```
celdrake, line 7: We'd need to make changes to the "Add devices" modal if we're not supporting any option for adding devices, and checkout the UX with Nir.

And ideally also remove the references to the two properties in the rest of the codebase (proxy, AppContext, etc)

apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/templates/flightctl-ui-deployment.yaml
```diff
@@ -22,34 +22,20 @@ spec:
...
                 key: FLIGHTCTL_SERVER
-          - name: FLIGHTCTL_METRICS_SERVER
-            valueFrom:
-              secretKeyRef:
-                name: flightctl-ui
-                key: FLIGHTCTL_METRICS_SERVER
           - name: FLIGHTCTL_GRPC_SERVER
             valueFrom:
-              secretKeyRef:
+              configMapKeyRef:
```
mangelajo: 👍

apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/values.yaml
mangelajo: good catch, it's 7444

apps/standalone/deploy/helm/flightctl-ui/Chart.yaml
mangelajo: 1. When we do the helm publish via an OCI container, it's injected into the Chart.yaml

apps/standalone/deploy/helm/flightctl-ui/templates/flightctl-ui-certs-secret.yaml
```diff
@@ -2,7 +2,8 @@ apiVersion: v1
...
 metadata:
   name: flightctl-ui-certs
-  namespace: {{ .Values.flightctlUi.namespace }}
+  namespace: {{ .Release.Namespace }}
 type: Opaque
 stringData:
-  ca.crt: {{ .Values.flightctlUi.certs.ca | quote }}
\ No newline at end of file
+  ca.crt: {{ .Values.api.caCert | quote }}
+  ca_oidc.crt: {{ default (.Values.global).flightctl.auth.caCert .Values.auth.caCert | quote }}
```
mangelajo:
```suggestion
  ca_oidc.crt: {{ default (.Values.global).flightctl.auth.caCert .Values.auth.caCert | quote }}
```
Just curious, why the braces? are those mandatory?

apps/standalone/deploy/helm/flightctl-ui/values.yaml
```diff
@@ -1,23 +1,26 @@
...
+  pullPolicy: Always
+## @param baseURL Base URL of UI.
+baseURL: ""
+## @param auth.internalOidcAuthority URL of OIDC authority that is accessible by UI backend.
+## @param auth.insecureSkipTlsVerify Set to true if OIDC TLS certificate validation should be skipped.
+## @param auth.caCert A custom CA cert for OIDC TLS.
+auth:
+  internalOidcAuthority: ""
+  insecureSkipTlsVerify: false
+  caCert: ""
```
mangelajo:
```suggestion
auth:
  internalOidcAuthority: ""
  insecureSkipTlsVerify: false
  caCert: ""
```

We should converge to put the auth details as globals, because I see the API also uses those.

https://github.com/flightctl/flightctl/blob/main/deploy/helm/flightctl/values.yaml#L65

What about having all of them as global.flightctl.auth   ?

apps/standalone/deploy/helm/flightctl-ui/values.yaml
```diff
@@ -1,23 +1,26 @@
...
-    ca: |
-      -----BEGIN
-      -----END
-    frontRouter: |
-      -----BEGIN
-      -----END
-    frontRouterKey: |
-      -----BEGIN
-      -----END
+## @param image.image Image to deploy.
```
mangelajo: We should probably document any global parameters being consumed.

apps/standalone/deploy/helm/flightctl-ui/values.yaml
```diff
@@ -1,23 +1,26 @@
...
+  internalOidcAuthority: ""
+  insecureSkipTlsVerify: false
+  caCert: ""
+## @param api.url URL of Flightctl API endpoint accessible by UI backend.
+## @param api.grpcUrl URL of Flightctl GRPC endpoint accessible by UI backend.
+## @param api.insecureSkipTlsVerify Set to true if Flightctl TLS certificate validation should be skipped.
+## @param api.caCert A custom CA cert for Flightctl TLS.
+api:
+  url: https://flightctl-api:3443/
+  grpcUrl: grpcs://flightctl-api-agent-grpc:7443/
```
mangelajo:
```suggestion
  grpcUrl: grpcs://flightctl-api-agent-grpc:7444/
```

apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/values.yaml
```diff
@@ -1,12 +1,17 @@
...
-  qcow2ImgUrl: "" #Qcow2 device image. Example: https://example.com/disk.qcow2
-  certs:
-    ca: |
-      -----BEGIN
-      -----END
+## @param image.image Image to deploy.
```
mangelajo: let's document also any global consumed values.

apps/standalone/deploy/helm/flightctl-ui/values.yaml
rawagner: Yes, I expect that we will provide these through globals. See that in the template I am using 
```
ca_oidc.crt: {{ default (.Values.global).flightctl.auth.caCert .Values.auth.caCert | quote }}
```

but I also add these in separate UI auth object in order to allow specifying different auth config than the global one. 
As maybe globally we want to use some `internalOidcAuthority` but for UI we want a different one. Im not sure why, but I think we want this flexibility.  WDYT ?

apps/standalone/deploy/helm/flightctl-ui/templates/flightctl-ui-certs-secret.yaml
rawagner: if `global` object is not defined we wont fail here. Without braces, the helm template fails with nil pointer exception

apps/standalone/deploy/helm/flightctl-ui/values.yaml
mangelajo: That would work, we just need to document the globals then :)

The bad side of adding too many knobs like this (global vs local) is that we need to maintain them later, I was considering on removing a lot of the knobs on the backend may be, but had not decided. 

What do you think? is there a possibility that we may need different oidc from the UI compared to the backend?

rawagner: Yeah, the bad side here is that we need to maintain these ..
Lets imagine a case where you deploy
 - FCLT API and Keycloak to namespace FOO
 - FCTL UI to namespace BAR
 
 then you could set `global.auth.internalOIDC` to `https://keycloak` and UI auth will need to be `https://keycloak.FOO`

or you could set global to `https://keycloak.FOO` and it will work for both  ...
(but for this we would need to enable deploying UI to a different NS than API)

I guess it mostly depends on what we agree :)

mangelajo: Yeah, good thinking,

It would make sense to agree to document that when a global.internalOidcAuthority is set, we must provide the full cluster path for the URL, including the namespace, and the same for any other internal endpoints. Then, it may work in all cases. We should probably do it regardless of keeping the specific knobs.


Another question I have is if we could auto-calculate that knob value if keycloak is enabled in the deployment. May be that is just some further refinement we could think of.

apps/standalone/deploy/helm/flightctl-ui/templates/flightctl-ui-certs-secret.yaml
mangelajo: nice I didn't know about that.

apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/values.yaml
mangelajo: This one is still pending in the same way we have for the other values:

https://github.com/flightctl/flightctl-ui/pull/112/files#diff-9712335e7da43c32d9689c43790084c6645fdaa287062d9f2b4eac266dbc07e1R6

mangelajo: Or we don't need any of those at all?

rawagner: Not at the moment. Maybe later we will add some, but currently I dont see any that would be useful here.

mangelajo (approved):

celdrake:
@rawagner The fix for the Cypress tests is https://github.com/flightctl/flightctl-ui/pull/115/commits/59eefdb3514af53a9b1f424755472cd8f9928c88. 

I can extract it to a separate PR if you prefer, so we can merge this PR.

mangelajo:
> @rawagner The fix for the Cypress tests is [59eefdb](https://github.com/flightctl/flightctl-ui/commit/59eefdb3514af53a9b1f424755472cd8f9928c88).
> 
> I can extract it to a separate PR if you prefer, so we can merge this PR.

The tests aren't mandatory yet. So you could technically merge them individually, in the end you know what you are doing and this PR didn't break the integration tests I believe.

celdrake:
> > @rawagner The fix for the Cypress tests is [59eefdb](https://github.com/flightctl/flightctl-ui/commit/59eefdb3514af53a9b1f424755472cd8f9928c88).
> > I can extract it to a separate PR if you prefer, so we can merge this PR.
> 
> The tests aren't mandatory yet. So you could technically merge them individually, in the end you know what you are doing and this PR didn't break the integration tests I believe.

Well, the merge button does not become enabled unless the CI step passes. The tests were not broken in this PR, so I'd rather fix them so we can get green builds and merge necessary PRs. I'm doing this now here https://github.com/flightctl/flightctl-ui/pull/117

rawagner:
Thanks @celdrake for fixing the tests. I will mark this as a draft as I'd like to get an agreement on the backend charts before merging this. I will be opening the PR for backend shortly

rawagner:
Backend PR https://github.com/flightctl/flightctl/pull/554

mangelajo:
@rawagner is this still draft? I am reviewing.

deploy/helm/flightctl-ui/templates/flightctl-ui-deployment.yaml
```diff
@@ -20,61 +20,66 @@ spec:
...
-          image: {{ .Values.flightctlUi.image }}
+            valueFrom:
+              configMapKeyRef:
+                name: flightctl-ui
+                key: OIDC_INSECURE_SKIP_VERIFY
+          {{- end }}
+          {{- if .Values.image.image }}
+          image: {{ .Values.image.image }}:{{ default .Chart.AppVersion .Values.image.tag }}
+          {{- else if eq (.Values.global).target "acm" }}
+          image: quay.io/flightctl/flightctl-ocp-ui:{{ default .Chart.AppVersion .Values.image.tag }}
```
mangelajo, line 67: 🎸

deploy/helm/flightctl-ui/templates/flightctl-ui-service.yaml
```diff
@@ -2,11 +2,13 @@ kind: Service
 apiVersion: v1
 metadata:
   name: flightctl-ui
-  namespace: {{ .Values.flightctlUi.namespace }}
+  namespace: {{ .Release.Namespace }}
   labels:
     app: flightctl-ui
+  {{- if not ((.Values.global).baseDomainTls).cert }}
```
mangelajo, line 8:
```suggestion
  {{- if not ((.Values.global).baseDomainTls).cert }}
```

Why is this inverted everywhere, shouldn't it be:

  {{- if ((.Values.global).baseDomainTls).cert }}

?

May be I just need more coffee ':D

mangelajo: also confused with the instances in this file : https://github.com/flightctl/flightctl-ui/pull/112/files#diff-325541283d9f28d1334be3088b93999822f1624d9b7a435c54c76c6aa4a7a564R79

rawagner: This condition means:
if user DOES NOT specify their own TLS cert (via `global.baseDomainTls`), we will annotate the Service and let OpenShift create a cert for us.

rawagner: You're right about the second one, the condition should not be there at all.

mangelajo (approved):
Looks good to me :)
time to try! :D

PR #115, opened 2024-09-23
EDM-470: Split fleets and resources syncs in FleetsPage

celdrake:
The FleetsPage will now include both Fleets and Resource syncs, but they will not be part of the same table.
Instead, pending RSs are now displayed as notifications at the top of the page, and users can dismiss them.

- The Fleet list columns have changed: "device selector" is removed, and instead we'll have the number of devices bound to each fleet.

rs-fleets

**Notes**: 
- Until https://issues.redhat.com/browse/MGMT-17879 is complete, we need to filter the RSs in the client side. The current filter will return all RS in an error state, and those that don't have a corresponding fleet yet. 

- We will dismiss RSs based on their "generation", since it seems to only change when a modification is done in the RS repository. Otherwise if using "lastTransitionTime", we would get different items every ~2 minutes based on my tests.


celdrake:
After making the necessary adaptations to the Cypress tests due to the new changes, there was a flake in the "Create fleet" test. Re-running fixed it, however I'll take a deeper look if it starts failing intermitently.

rawagner (approved):

PR #116, opened 2024-09-30
Fix frontend access to UI on OCP plugin mode

mangelajo:
Also make the network policy more specific to target only the UI pods, otherwise on deployments where the "internal" and "external" services live on the same namespace, the policy was also providing access to DB/rabbitmq/etc.

Related: https://issues.redhat.com/browse/EDM-463

### DIFF
--- a/apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/templates/flightctl-ui-network-policy.yaml
+++ b/apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/templates/flightctl-ui-network-policy.yaml
@@ -1,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: flightctl-ui-from-ingress 
+  namespace:  {{ .Release.Namespace }}
+spec:
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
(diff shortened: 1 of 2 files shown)


apps/ocp-plugin/deploy/helm/flightctl-ocp-ui/templates/flightctl-ui-network-policy.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
```
celdrake, line 1: Shouldn't we consider the "standalone" deployment the base one, and the OCP one the overwrites?

Eg. if we had a different plugin, it would extend the OCP one which feels a bit weird.

What I mean is to define the NP for the standalone deployments, then have the OCP plugin import it.

rawagner (changes requested):
We need different policies for Standalone and OCP plugin.
For Standalone we need to allow external access. So your existing `flightctl-ui-network-policy` works fine (but should be moved to Standalone helm chart).

For OCP plugin, there is no route/external access. Instead there is only Service which needs to be accessible by OCP console pod which is in `openshift-console` namespace.

rawagner:
I will close this as it is outdated and I believe the issue is already fixed by plenty of other changes we did previously :) Feel free to reopen if needed

PR #129, opened 2024-10-09
NO-ISSUE: Add /enroll route to ACM plugin

rawagner:


oglok:
LGTM, it's required for visualizing ER URLs created by the agent.
Thanks @rawagner

PR #132, opened 2024-10-15
EDM-461: Integrate with new Inline Configs API

rawagner:
inline_upd


libs/ui-components/src/components/Device/EditDeviceWizard/deviceSpecUtils.ts
```diff
@@ -59,7 +57,23 @@ const isSameSecretConf = (a: KubernetesSecretProviderSpec, b: KubernetesSecretPr
...
 
 const isSameInlineConf = (a: InlineConfigProviderSpec, b: InlineConfigProviderSpec) => {
-  return a.name === b.name && isEqual(a.inline, b.inline);
+  return (
+    a.name === b.name &&
+    a.inline.length === b.inline.length &&
+    a.inline.every((aInline, index) => {
+      const bInline = b.inline[index];
+      return (
+        aInline.content === bInline.content &&
```
celdrake: The content field would be the more expensive comparison and it would be best to compare it last.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
```diff
@@ -1,20 +1,152 @@
...
+  const permissions = React.useMemo(
+    () => ({
+      '777': t('(0777) Read, write, and execute permissions for all users.'),
+      '755': t("(0755) Read and execute permission for all users. The file's owner also has write permission."),
+      '750': t(
+        "(0750) Read and execute permission for the owner and group. The file's owner also has write permission. Users who aren't the file's owner or members of the group have no access to the file.",
+      ),
+      '700': t("(0700) Read, write, and execute permissions for the file's owner only; all others have no access."),
+      '666': t('(0666) Read and write permissions for all users. No execute permissions for anybody.'),
+      '664': t('(0664) Read and write permissions for the owner and group. Read- only permission for all others.'),
```
celdrake:
```suggestion
      '664': t('(0664) Read and write permissions for the owner and group. Read-only permission for all others.'),
```

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
```diff
@@ -1,20 +1,152 @@
...
+import { DeviceSpecConfigFormValues } from '../types';
+import { InlineConfigTemplate } from '../../../../types/deviceSpec';
+import ExpandableFormSection from '../../../form/ExpandableFormSection';
+import SwitchField from '../../../form/SwitchField';
+
+const FileForm = ({ fieldName }: { fieldName: string }) => {
+  const { t } = useTranslation();
+  const [{ value: file }] = useField<InlineConfigTemplate['files'][0]>(fieldName);
+
+  const permissions = React.useMemo(
```
celdrake, line 21: I would try to be more consistent in the wording.

eg. "All others have no access, no world permissions, no permissions for others, no permission for anybody else" all mean the same

libs/ui-components/src/components/form/FormSelectTypeahead.tsx
```diff
@@ -0,0 +1,199 @@
...
+  children,
+  validateNewItem,
+  optionExists,
+}) => {
+  const { t } = useTranslation();
+  const [field, meta, { setValue, setTouched }] = useField<string>({
+    name: name,
+  });
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [inputValue, setInputValue] = React.useState<string | undefined>(undefined);
```
celdrake, line 50: When we start typing to search or create an item, the existing text is not cleared.
It happens with both the placeholder text, as well with the existing selection.

IMO when users start typing, it shouldn't be necessary to clear the existing value first.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
```diff
@@ -1,20 +1,152 @@
...
+          <FormSelectTypeahead
+            name={`${fieldName}.permissions`}
+            placeholderText={t('Select permissions')}
+            items={permissions}
+            validateNewItem={(value) => {
+              try {
+                if (value !== '' && value.length <= 4) {
+                  const valNum = Number(`0o${value}`);
+                  if (isFinite(valNum) && valNum >= 0 && valNum <= 0o7777) {
+                    return undefined;
```
celdrake, line 61: Are we allowing to define both octal and decimal permissions? In that case, we probably should make a distinction in the UI. (eg show something like (0345 / Decimal 229)).

Screenshot from 2024-10-15 11-42-58

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
```diff
@@ -1,20 +1,152 @@
...
+              } catch {
+                return t('Invalid permissions');
+              }
+              return t('Invalid permissions');
+            }}
+            optionExists={(value) => Object.keys(permissions).some((k) => k === value || `0${k}` === value)}
+          />
+        </FormGroup>
+        <FormGroup label={t('User')}>
+          <TextField name={`${fieldName}.user`} />
```
celdrake: We should either add a placeholder or a help text icon indicating that the default is "root"
Same for the group.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
```diff
@@ -1,20 +1,152 @@
...
+          <TextField name={`${fieldName}.path`} />
+        </FormGroup>
+        <FormGroup label={t('Content')} isRequired>
+          <TextAreaField name={`${fieldName}.content`} />
+        </FormGroup>
+        <SwitchField label={t('Content is base64 encoded')} name={`${fieldName}.base64`} />
+        <FormGroup label={t('Permissions')}>
+          <FormSelectTypeahead
+            name={`${fieldName}.permissions`}
+            placeholderText={t('Select permissions')}
```
celdrake, line 53: According to the design, there should be a permanent helper text as well, that explains that permissions can be either selected or typed.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
```diff
@@ -1,20 +1,152 @@
...
+  return (
+    <ExpandableFormSection title={t('File')} fieldName={fieldName} description={file.path}>
+      <Grid hasGutter>
+        <FormGroup label={t('File path on the device')} isRequired>
+          <TextField name={`${fieldName}.path`} />
+        </FormGroup>
+        <FormGroup label={t('Content')} isRequired>
+          <TextAreaField name={`${fieldName}.content`} />
+        </FormGroup>
+        <SwitchField label={t('Content is base64 encoded')} name={`${fieldName}.base64`} />
```
celdrake, line 49: Switching the toggle for file 2 updates toggles the switch in file 1 instead.

celdrake (changes requested):

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: No, we allow only octal notation. I dont like supporting both as it just makes it too confusing.

celdrake: I agree we shouldn't mix.
In that case, the `validateItem`  function doesn't validate correctly the values, because as in the screenshot, it accepts values like "44". We shouldn't allow creating items unless it's a valid octal value.

celdrake: As discussed, this would be the octal value "44" but showing it like that instead of with the octal notation as for the other entries can be confusing. 
I suggest allowing valid octal values such as  "44"l, but then IMO we should then show them in the same format "0044" as the other items, when it is added to the Dropdown.

libs/ui-components/src/components/Device/EditDeviceWizard/deviceSpecUtils.ts
rawagner: Ok, moved as last item

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: fixed

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: I've actually removed these. We just focus on what is allowed.

libs/ui-components/src/components/form/FormSelectTypeahead.tsx
rawagner: Should be fixed now

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: user can type 44 still, but the `Add new item` will show 0044. The user's input is always transformed to 4 digit length

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: placeholder added for user and group too

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: I cannot reproduce..maybe I fixed it in some other changes ? Can you try again please ?

celdrake: Yes, it's fixed.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
rawagner: added

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ConfigInlineTemplateForm.tsx
celdrake: :+1: that prevents the confusion

celdrake (approved):

PR #140, opened 2024-10-24
EDM-615: Loading DescriptionList CSS from PF5 explicitly

celdrake:
Alternative workaround to load the PF5 styles for DescriptionList in the OCP plugin.


celdrake:
@rawagner I merged the other PR that used the PF4 class to avoid loading extra PF classes besides Console.

PR #151, opened 2024-11-07
EDM-630: New design for device template step

celdrake:
The device template step UX has changed to give a better separation between Configurations and Applications.

Screenshot from 2024-11-07 15-17-54


celdrake:
Added "do-no-merge" as only bugfixes should be merged until v0.3.0 is released.

rawagner (approved):

PR #155, opened 2024-11-20
EDM-472: Sorting entities done via the API only

celdrake:
Removes all sorting done client-side. The API currently supports only sorting by metadata.name and only in Ascending order.

Therefore, in the UI we remove the functionality to be able to sort by several columns, or to change the sorting direction in the metadata.name column.

We need to explicitly indicate the default sorting to all entities, to prevent a bug which happens when we request for paginated results (which we will do following this PR).


celdrake:
Marking as Draft as there are discussions about maybe not being able to sort via the API by fields other than the default one (eg. metadata.name)

rawagner (approved):

PR #156, opened 2024-11-21
EDM-473 Fleets table includes pagination

celdrake:
Adds pagination to Fleets table. If the approach is approved, we can extend this pattern to the rest of the tables.

Also, it changes the logic on how we show the resource syncs, so it doesn't depend in fetching the entire list of fleets.

fleets-paginated

### DIFF
--- a/libs/ui-components/src/components/Table/TablePagination.tsx
+++ b/libs/ui-components/src/components/Table/TablePagination.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { Trans } from 'react-i18next';
+import { Pagination, PaginationToggleTemplateProps, PaginationVariant, Spinner } from '@patternfly/react-core';
+
+import { PAGE_SIZE } from '../../constants';
+import { useTranslation } from '../../hooks/useTranslation';
+
+type TablePaginationProps = {
+  isUpdating: boolean;
+  itemCount?: number;
+  currentPage: number;
+  setCurrentPage: (page: number) => void;
+};
+
+const PaginationTemplate = ({
+  isUpdating,
+  firstIndex,
+  itemCount,
+}: PaginationToggleTemplateProps & { isUpdating: boolean }) => {
+  const { t } = useTranslation();
+
+  if (isUpdating) {
+    return <Spinner size="sm" />;
+  }
+
+  const totalPages = String(Math.ceil((itemCount || 0) / PAGE_SIZE));
+  const currentPage = String(Math.ceil((firstIndex || 0) / PAGE_SIZE));
+  return (
+    <Trans t={t}>
+      {currentPage} of <strong>{totalPages}</strong>
+    </Trans>
+  );
+};
+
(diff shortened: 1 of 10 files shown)


libs/ui-components/src/hooks/useTablePagination.ts
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { ApiList } from '../utils/api';
+import { PAGE_SIZE } from '../constants';
+
+export const useTablePagination = () => {
+  const [currentPage, setCurrentPage] = React.useState<number>(1);
+  const [continueTokens, setContinueTokens] = React.useState<string[]>([]);
```
rawagner, line 7: storing previous tokens isn't optimal because as soon as the list starts changing a bit, they will be pointing to incorrect places.

But I understand we currently do not have a proper support from backend to go back so this is the only way.
We should however ask for supporting a back tokens.

That way you'd only look at the current list results. First item would be your back token and the last item would be your continue token (as is now).

rawagner: Not a blocker for this PR

libs/ui-components/src/components/Table/TablePagination.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const TablePagination = ({ isUpdating, itemCount, currentPage, setCurrentPage }: TablePaginationProps) => {
+  const { t } = useTranslation();
+
+  const onSetPage = (_event: React.MouseEvent | React.KeyboardEvent | MouseEvent, newPage: number) => {
+    setCurrentPage(newPage);
+  };
+
+  return (
+    <Pagination
```
rawagner, line 56: clicking fast on back/forward messes up the pagination. It won't change the content of the page. Maybe we should disable these buttons while fetching ?

rawagner (changes requested):

libs/ui-components/src/hooks/useTablePagination.ts
celdrake: Yes, it's as you said. For going back, the risk is that the token for the previous page may be outdated. 
For moving forward, we update the token based on the last request for the current page, so that should work well.

I'll create a story for Backend to provide the back token as well.

libs/ui-components/src/components/Table/TablePagination.tsx
celdrake: Good idea.

libs/ui-components/src/hooks/useTablePagination.ts
celdrake: Created https://issues.redhat.com/browse/EDM-844 for backend and EDM-845 for UI.

libs/ui-components/src/components/Table/TablePagination.tsx
celdrake: I also added an improvement to avoid showing only the spinner while the new page is loading.
Now we'll show the loader and also the page number control.

rawagner (approved):

PR #157, opened 2024-11-21
EDM-747: Create Ansible lib

rawagner:


libs/ui-components/src/components/common/LeaveFormConfirmation.tsx
```diff
@@ -79,11 +79,16 @@ const BrowserBlocker = ({ lock }: { lock: boolean }) => {
 const LeaveFormConfirmation = () => {
   const { dirty, isSubmitting } = useFormikContext();
   const {
+    appType,
     router: { useBlocker },
   } = useAppContext();
 
   const lock = !isSubmitting && dirty;
 
+  if (appType === 'aap') {
```
rawagner, line 88: even though AAP UI uses new react-router, we cannot (yet) use the `useBlocker` hook as they still use `BrowserRouter` and not data router.

libs/ansible/src/hooks/useFetch.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const patch = React.useCallback(
+    async <R>(kind: string, obj: PatchRequest, abortSignal?: AbortSignal): Promise<R> =>
+      patchData(kind, obj, serviceUrl, applyHeaders, abortSignal),
+    [serviceUrl, applyHeaders],
+  );
+
+  const getWsEndpoint = React.useCallback(
+    () => ({
+      wsEndpoint: '',
```
rawagner, line 45: no ws yet

libs/ui-components/package.json
```diff
@@ -36,10 +47,10 @@
     "@patternfly/react-icons": "^5.0.0",
     "@patternfly/react-styles": "^5.1.2",
     "@patternfly/react-table": "^5.0.0",
-    "i18next": "^21.8.14",
-    "react": "17.0.1",
-    "react-dom": "17.0.1",
-    "react-i18next": "^11.7.3",
+    "i18next": ">=21.8.14",
```
celdrake: Should these versions be perhaps more limited?
What versions are we targeting?

libs/ansible/src/const.ts
```diff
@@ -0,0 +1,20 @@
...
+  [ROUTE.FLEET_EDIT]: '/edge/fleets/edit',
+  [ROUTE.FLEET_IMPORT]: '/edge/fleets/import',
+  [ROUTE.DEVICES]: '/edge/devices',
+  [ROUTE.DEVICE_DETAILS]: '/edge/devices',
+  [ROUTE.DEVICE_EDIT]: '/edge/devices/edit',
+  [ROUTE.REPO_CREATE]: '/edge/repositories/create',
+  [ROUTE.REPO_EDIT]: '/edge/repositories/edit',
+  [ROUTE.REPO_DETAILS]: '/edge/repositories',
+  [ROUTE.REPOSITORIES]: '/edge/repositories',
+  [ROUTE.RESOURCE_SYNC_DETAILS]: '/edge/resourcesyncs',
```
celdrake, line 17: I just realised that in `useNavigate` the Routes for Repository details, Fleet details and RS details contain an extra space.

Could you change them in this PR, please?

libs/ansible/src/hooks/useFetch.ts
celdrake: Should we avoid showing the Terminal tab in that case?

celdrake (commented):
Great work, just a few small comments

libs/ui-components/package.json
rawagner: i18next 23.15.2
react 18.3.1
react-dom 18.3.1
react-i18next 15.0.2

I've updated the ranges. Let me know if you want them restricted even more.

libs/ansible/src/const.ts
rawagner: Good catch, fixed

libs/ansible/src/hooks/useFetch.ts
rawagner: yep, i've added a prop to hide the tab and it is being used from ansible lib

libs/ui-components/package.json
celdrake: Looks good now, thanks

celdrake (approved):

PR #158, opened 2024-11-25
Add release job to release fctl libs from release branches

rawagner:
Mostly a copy-paste from the assisted-installer-ui repo.
Just changed the package names & replaced yarn with npm (which has slightly different commands)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,90 @@
+name: Release a new version
+
+on:
+  release:
+    types:
+      - published
+
+permissions:
+  contents: write
+
+env:
+  NODE_OPTIONS: '--max-old-space-size=8192'
+
+jobs:
+  publish-to-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check workflow runs in a release branch
+        run: |
+          if [[ "${{ github.event.release.target_commitish }}" != *releases* ]]; then
+            echo "Expecting to be in a release branch, but we are in: ${{ github.event.release.target_commitish }}"
+            exit 1
+          fi
+
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.release.target_commitish }}
+          fetch-depth: 0
+
+      - name: Lookup for workspaces to be published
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,88 @@
...
+env:
+  NODE_OPTIONS: '--max-old-space-size=8192'
+
+jobs:
+  publish-to-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check workflow runs in a release branch
+        run: |
+          if [[ "${{ github.event.release.target_commitish }}" != *releases* ]]; then
```
celdrake, line 20: This workflow requires a release branch.

IIRC, this means we'll need to create the tag by targeting a release branch rather than `main`.
Is there anything else that we'll need to do?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,88 @@
...
+            echo "Expecting to be in a release branch, but we are in: ${{ github.event.release.target_commitish }}"
+            exit 1
+          fi
+
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.PUSH_TO_PROTECTED_BRANCH }}
```
celdrake: We don't have this secret in the settings ATM.

rawagner: I removed this one and instead used 
```
permissions:
  contents: write
```

hard to test though :)

.github/workflows/release.yaml
rawagner: you're correct, we would create a release branch and then we would create a new release via Github releases. It actually creates a tag.

.github/workflows/release.yaml
celdrake: Yes, we'll need to do an actual release to test it for real.
But the change looks good.

.github/workflows/release.yaml
celdrake: Nice, it's not so different to the current flow.

celdrake (approved):

PR #166, opened 2024-12-03
EDM-766: Add MicroShift registration checkbox

rawagner:


libs/ui-components/src/utils/patch.ts
```diff
@@ -200,3 +206,54 @@ export const getApplicationPatches = (
...
+      patches.push({
+        op: 'add',
+        path: `${basePath}/hooks/afterUpdating`,
+        value: [ACMCrdHook, ACMImportHook],
+      });
+    } else {
+      if (!currentHooks.afterUpdating?.some(isAcmCrdHook)) {
+        patches.push({
+          op: 'add',
+          path: `${basePath}/hooks/afterUpdating/-`,
```
celdrake: Seems like a typo?

```suggestion
          path: `${basePath}/hooks/afterUpdating/`,
```

libs/ui-components/src/utils/patch.ts
```diff
@@ -200,3 +206,54 @@ export const getApplicationPatches = (
...
+          value: ACMCrdHook,
+        });
+      }
+      if (!currentHooks.afterUpdating?.some(isAcmImportHook)) {
+        patches.push({
+          op: 'add',
+          path: `${basePath}/hooks/afterUpdating/-`,
```
celdrake: And here
```suggestion
          path: `${basePath}/hooks/afterUpdating/`,
```

libs/ui-components/src/components/Device/EditDeviceWizard/steps/DeviceTemplateStep.tsx
```diff
@@ -21,7 +25,62 @@ const templateOption1 = '{{ device.metadata.labels[key] }}';
 const templateOption2 = '{{ device.metadata.name }}';
 const exampleCode = `/device-configs/factory-floors/floor-{{ device.metadata.labels[factory-floor] }}`;
 
+const MicroShiftCheckbox = ({ isFleet }: { isFleet: boolean }) => {
+  const { t } = useTranslation();
+
+  const [repo, loading, error] = useFetchPeriodically<Required<Repository>>({
+    endpoint: `repositories/acm-registration`,
```
celdrake: Would be better to define the repository name in a constant.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/DeviceTemplateStep.tsx
```diff
@@ -21,7 +25,62 @@ const templateOption1 = '{{ device.metadata.labels[key] }}';
...
+                  {t(
+                    "To remove registration, you'll need to uncheck this option and also remove the cluster from ACM's clusters list",
+                  )}
+                </>
+              }
+            />
+          </>
+        )
+      }
+      isDisabled={!repo || !!error}
```
celdrake: IMO there should be an explanation that the checkbox is disabled because the repository `acm-registration` is missing.

That would help customers in case they deleted the repo etc, to put it back.

libs/ui-components/src/utils/patch.ts
```diff
@@ -200,3 +206,54 @@ export const getApplicationPatches = (
...
+  currentHooks: DeviceHooksSpec | undefined,
+  registerMicroShift: boolean,
+  basePath: string,
+) => {
+  const patches: PatchRequest = [];
+  if (registerMicroShift) {
+    if (!currentHooks) {
+      patches.push({
+        op: 'add',
+        path: `${basePath}/hooks/`,
```
celdrake: Seems to fail due to the trailing slash.

libs/ui-components/src/utils/patch.ts
```diff
@@ -200,3 +206,54 @@ export const getApplicationPatches = (
...
+      });
+    } else if (!currentHooks.afterUpdating) {
+      patches.push({
+        op: 'add',
+        path: `${basePath}/hooks/afterUpdating`,
+        value: [ACMCrdHook, ACMImportHook],
+      });
+    } else {
+      if (!currentHooks.afterUpdating?.some(isAcmCrdHook)) {
+        patches.push({
```
celdrake: I think the flow here is wrong, because I had an initial Git config, and then I checked the checkbox.
It attempted to add two separate patches to `/spec/template/spec/hooks/afterUpdating`, rather than setting it to an array which includes both hooks.


```
...other patch items...
{
    "op": "add",
    "path": "/spec/template/spec/hooks/afterUpdating",
    "value": {
      "path": "/var/local/acm-import/crd.yaml",
      "onFile": [
        "Create"
      ],
      "actions": [
        {
          "executable": {
            "envVars": [
              "KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig"
            ],
            "run": "kubectl apply -f /var/local/acm-import/crd.yaml"
          }
        }
      ]
    }
  },
  {
    "op": "add",
    "path": "/spec/template/spec/hooks/afterUpdating",
    "value": {
      "path": "/var/local/acm-import/import.yaml",
      "onFile": [
        "Create"
      ],
      "actions": [
        {
          "executable": {
            "envVars": [
              "KUBECONFIG=/var/lib/microshift/resources/kubeadmin/kubeconfig"
            ],
            "run": "kubectl apply -f /var/local/acm-import/import.yaml"
          }
        }
      ]
    }
  }
```

It results in an error 400 because of it.

libs/ui-components/src/utils/patch.ts
rawagner: this is not a typo. The `-` means that we want to put the value at the end of the array. See https://jsonpatch.com/#add

libs/ui-components/src/utils/patch.ts
rawagner: with the `/-` at the end, it should work. But you're right that I could optimize this and do it in one patch for some cases.

libs/ui-components/src/utils/patch.ts
rawagner: fixed

libs/ui-components/src/components/Device/EditDeviceWizard/steps/DeviceTemplateStep.tsx
rawagner: constant is created, but not used here :) fixed

libs/ui-components/src/utils/patch.ts
celdrake: I didn't know that, thanks!

libs/ui-components/src/utils/patch.ts
rawagner: fixed

libs/types/models/ConditionType.ts
```diff
@@ -4,17 +4,14 @@
...
+  CertificateSigningRequestApproved = 'Denied',
+  CertificateSigningRequestDenied = 'Failed',
+  CertificateSigningRequestFailed = 'Accessible',
+  RepositoryAccessible = 'ResourceParsed',
+  ResourceSyncAccessible = 'Synced',
+  ResourceSyncResourceParsed = 'OverlappingSelectors',
+  ResourceSyncSynced = 'Valid',
+  FleetOverlappingSelectors = 'Updating',
+  FleetValid = 'SpecValid',
+  DeviceUpdating = 'MultipleOwners',
```
rawagner: These seem broken completely, WDYT @celdrake  ?

celdrake: Yes, since the API defines duplicate keys, the enum shifts.

The only solution I found is to keep reverting the file, and only update it when there's a change in the API definitions (requires making manual changes).

I checked the generator back in the day and didn't found a way to fix the problem from root.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/DeviceTemplateStep.tsx
rawagner: Ok, I've added an alert below with link to docs on how to re-create the repo if needed.

celdrake: Looks great, thanks!

celdrake (approved):
Works great!
I'm just wondering if the Fleet details page should indicate that the ACM registration is enabled. ATM we would display the configurations we created automatically, but it may not be obvious to the user what that means. Could be addressed in a follow-up.

libs/ui-components/src/components/Device/EditDeviceWizard/deviceSpecUtils.ts
```diff
@@ -224,55 +246,136 @@ export const getApplicationValues = (deviceSpec?: DeviceSpec): ApplicationFormSp
...
+
+const isConfigACMImport = (c: ConfigProviderSpec) => {
+  if (!isHttpProviderSpec(c)) {
+    return false;
+  }
+  return (
+    c.name === ACMImportConfig.name &&
+    c.httpRef.filePath === ACMImportConfig.httpRef.filePath &&
+    c.httpRef.repository === ACMImportConfig.httpRef.repository &&
+    c.httpRef.suffix === ACMImportConfig.httpRef.suffix
```
celdrake: Looks like when changing a device labels so that it's not part of a fleet anymore, the config gets evaluated and the `metadata.name` part is replaced by the actual value.

Maybe it's best to assume that the config with the same name must necessarily be of the expected content?
As there can only be 1 config with such a name. (or at least, skip the part with the variables)

rawagner: great catch, fixed

PR #170, opened 2024-12-09
EDM-624: Sorting parameters have been removed from the API

celdrake:
Due to bug EDM-624, the API parameters `sortBy` and `sortOrder` have been removed for now.

We'll have a proper sorting implementation in the future, but for now the default sorting is for `metadata.name` in Ascending order, and it will be done implicitly in the API.


celdrake:
Merging as it's needed for other PRs.

PR #172, opened 2024-12-11
EDM-855 template variables use go syntax

celdrake:
Small changes in the UX as the template variables syntax is now more complex.


rawagner (changes requested):
can you please update the Microshift template too ?

rawagner (approved):

PR #174, opened 2024-12-11
EDM-847: Add RBAC checks

rawagner:
This is an initial, low-scale implementation of checking if user has permissions to do some action (create/import fleet). If we agree that the approach and implementation is ok, I will expand to other parts of UI.


libs/ui-components/src/hooks/useAppContext.tsx
```diff
@@ -34,6 +34,19 @@ export const appRoutes = {
   [ROUTE.ENROLLMENT_REQUEST_DETAILS]: '/devicemanagement/enrollmentrequests',
 };
 
+export enum VERB {
+  CREATE = 'create',
```
celdrake: We'll also need support for `patch`, and maybe also `update` (for Decommissioning devices).

libs/ui-components/src/components/Fleet/FleetsPage.tsx
```diff
@@ -34,19 +34,26 @@ import { ROUTE, useNavigate } from '../../hooks/useNavigate';
...
 import FleetResourceSyncs from './FleetResourceSyncs';
 import { useFleetBackendFilters, useFleets } from './useFleets';
+import { useAccessReview } from '../../hooks/useAccessReview';
+import { RESOURCE, VERB } from '../../hooks/useAppContext';
 
 const FleetPageActions = ({ createText }: { createText?: string }) => {
   const { t } = useTranslation();
   const navigate = useNavigate();
+  const [isCreateAllowed, isCreateLoading, createCheckError] = useAccessReview(RESOURCE.FLEET, VERB.CREATE);
+  const [isImportAllowed, isImportLoading, importCheckError] = useAccessReview(RESOURCE.REPOSITORY, VERB.CREATE);
```
celdrake: For importing a fleet, the permission to create a repository is optional (depends if they want to use an existing or new repository), but the permission to create a RS is always needed.

If at least 1 repo exists, they'd only need the RS permission, otherwise they'd need both. 
Maybe we can handle this as a follow-up, as I think the approach is great, just that this is a more complex case.

libs/ui-components/src/components/Fleet/FleetsPage.tsx
```diff
@@ -34,19 +34,26 @@ import { ROUTE, useNavigate } from '../../hooks/useNavigate';
...
+
+  const isCreateDisabled = !isCreateAllowed || isCreateLoading || !!createCheckError;
+  const isImportDisabled = !isImportAllowed || isImportLoading || !!importCheckError;
   return (
     <Split hasGutter>
       <SplitItem>
-        <Button variant="primary" onClick={() => navigate(ROUTE.FLEET_CREATE)}>
+        <Button variant="primary" onClick={() => navigate(ROUTE.FLEET_CREATE)} isDisabled={isCreateDisabled}>
```
celdrake: Nit: I'm wondering if it would be a good idea to create a component for the button + the access check.
We may want to include a message for the disabled buttons, and that would help DRY.

libs/ui-components/src/hooks/useAppContext.tsx
rawagner: yes, we will have even more (like approve ER, access console etc..), but these verbs are not finalized, so I put just some examples here for now.

libs/ui-components/src/components/Fleet/FleetsPage.tsx
rawagner: Ok, I changed the checks to allow accessing wizard if user has permissions for Create RS and List Repo.
In the wizard hiding the option to create a new repo if they dont have permissions to create.

rawagner:
Im quite sure I didn't catch everything, but should be ok for first pass.

libs/ui-components/src/components/Fleet/FleetsPage.tsx
rawagner: FYI after discussion with Nir, we decided to hide the elements instead of disabling them.

libs/ui-components/src/components/Device/DevicesPage/DevicesPage.tsx
```diff
@@ -30,20 +30,28 @@ import {
...
   return (
     <ResourceListEmptyState icon={MicrochipIcon} titleText={t('No devices here!')}>
       <EmptyStateBody>
-        <Trans t={t}>
-          You can add devices and label them to match fleets, or your can{' '}
-          <Link to={ROUTE.FLEET_CREATE}>start with a fleet</Link> and add devices into it.
-        </Trans>
+        {canCreateFleet ? (
+          <Trans t={t}>
+            You can add devices and label them to match fleets, or your can{' '}
```
celdrake: There was a typo here
```suggestion
            You can add devices and label them to match fleets, or you can{' '}
```

libs/ui-components/src/components/Device/DevicesPage/DevicesPage.tsx
```diff
@@ -189,7 +204,7 @@ export const DeviceTable = ({
   );
 };
 
-const DevicesPage = () => {
+const DevicesPage = ({ canListER }: { canListER: boolean }) => {
```
celdrake, line 207: When users can't list ERs, we need to disable the badge of pending devices and the "todo" card in the Overview page.

Can be done as a follow-up.

libs/ui-components/src/components/common/ButtonWithPermissions.tsx
```diff
@@ -0,0 +1,13 @@
+import { Button, ButtonProps } from '@patternfly/react-core';
+import * as React from 'react';
+
+type WithPermissions = {
+  permissions: [boolean, boolean, string | undefined];
```
celdrake: I don't understand this type well. The component is only using the first element?

libs/ui-components/src/components/Device/DeviceDetails/DeviceDetailsPage.tsx
```diff
@@ -29,6 +31,9 @@ const DeviceDetailsPage = ({ children, hideTerminal }: React.PropsWithChildren<{
 
   const deviceAlias = device?.metadata.labels?.alias || deviceId;
 
+  const [canDelete] = useAccessReview(RESOURCE.DEVICE, VERB.DELETE);
+  const [canEdit] = useAccessReview(RESOURCE.DEVICE, VERB.PATCH);
```
celdrake, line 35: There should also be a check in `EditDeviceWizard` to prevent the users from accessing the "edit" URL directly.
Can be done as a follow-up.

celdrake (approved):

libs/ui-components/src/components/common/ButtonWithPermissions.tsx
rawagner: For now, yes. I'm keeping this open in case in other UIs (like AAP), we will want to disable + maybe show some tooltip.

I could remove it, if you prefer

libs/ui-components/src/components/Device/DevicesPage/DevicesPage.tsx
rawagner: Oh, right. I completely forgot about Overview. I'll do it in a followup

libs/ui-components/src/components/Device/DeviceDetails/DeviceDetailsPage.tsx
rawagner: the same would apply for any other URL that can be accessed directly (edit fleet, repo etc.).
I can fix that in a followup.

libs/ui-components/src/components/common/ButtonWithPermissions.tsx
celdrake: what's the third parameter (second boolean) for?
We could keep it, but maybe add some description?

libs/ui-components/src/components/Device/DevicesPage/DevicesPage.tsx
rawagner: fixed

libs/ui-components/src/components/common/ButtonWithPermissions.tsx
rawagner: This is the result of `useAccessReview`. I've exported the type and used it here.

libs/ui-components/src/components/Device/DeviceDetails/DeviceDetailsPage.tsx
rawagner: The follow up is tracked in https://issues.redhat.com/browse/EDM-949

libs/ui-components/src/components/common/ButtonWithPermissions.tsx
celdrake: Thanks for the clarification!

PR #175, opened 2024-12-11
NO-ISSUE: Remove unnecessary PageSection from Devices page

rawagner:


celdrake:
According to the design, the page content should fill the viewport. And the other list pages all use PageSections.
I'm wondering why should this page be different?

rawagner:
> According to the design, the page content should fill the viewport. And the other list pages all use PageSections. I'm wondering why should this page be different?

as discussed offline, we already have PageSections here, even after removing the top one, since ListPage is a PageSection.
I tried ocp and standalone and haven't seen any issues.

celdrake:
Yes, I see it correctly now, IDK why I saw the content didn't fill the page earlier. Apologies!

celdrake (approved):

PR #176, opened 2024-12-12
EDM-883: Clear error after moving away from submit step

celdrake:
If the user attempts to make some changes in the Device/Fleet wizard, but it fails due to an API error, the error stayed visible even after the user moves to the previous steps and even if they fix the issue.

When the user moves to another step, the error should be cleared.

### DIFF
--- a/libs/ui-components/src/components/Device/EditDeviceWizard/EditDeviceWizard.tsx
+++ b/libs/ui-components/src/components/Device/EditDeviceWizard/EditDeviceWizard.tsx
@@ -69,7 +69,7 @@ const EditDeviceWizard = () => {
       <Formik<EditDeviceFormValues>
         initialValues={{
           deviceAlias,
-          osImage: device.spec?.os?.image,
+          osImage: device.spec?.os?.image || '',
           labels: fromAPILabel(device.metadata.labels || {}).filter((label) => label.key !== 'alias'),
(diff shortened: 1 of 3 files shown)


libs/ui-components/src/components/Device/EditDeviceWizard/EditDeviceWizard.tsx
```diff
@@ -69,7 +69,7 @@ const EditDeviceWizard = () => {
       <Formik<EditDeviceFormValues>
         initialValues={{
           deviceAlias,
-          osImage: device.spec?.os?.image,
+          osImage: device.spec?.os?.image || '',
```
celdrake, line 72: Avoids "uncontrolled input" React error in console when modifying the OS image.

libs/ui-components/src/components/Device/EditDeviceWizard/utils.ts
```diff
@@ -35,7 +35,8 @@ export const getDevicePatches = (currentDevice: Device, updatedDevice: EditDevic
 
   // Device labels
   const currentLabels = currentDevice.metadata.labels || {};
-  const updatedLabels = updatedDevice.labels || [];
+  const updatedLabels = [...updatedDevice.labels];
```
celdrake, line 38: Form values were being modified and the "alias" label would be duplicate when there was a failure in submitting the form.

rawagner (approved):

PR #179, opened 2024-12-18
Update types to latest as of Dec 18th 2024

celdrake:
Updates latest openapi spec.  Includes:

- Mostly changes to the field descriptions.
- A couple of new  fields related to device decommissioning.


celdrake:
Merged directly as the API changes in descriptions affected a very large number of files, and it's best to keep those separate from feature PRs.

PR #180, opened 2024-12-18
EDM-391: Allowing users to decommission devices from the UI

celdrake:
Includes all the changes related to Device actions:

- Edit device: it's not possible to edit a device after it has started decommissioning
- Delete device: it's not possible to delete a device unless it has started decommissioning already
- Decommission device: it's not possible to decommission a device that has already started decommissioning

I left parts of the DecommissionModal commented, as there will be additional `target` options in the future, although the initial implementation only allows to select `Unenroll`.

Needs review since https://github.com/flightctl/flightctl/pull/771 may not get merged in the end.


libs/ui-components/src/components/Device/DeviceDetails/DeviceDetailsTab.tsx
```diff
@@ -68,7 +68,11 @@ const DeviceDetailsTab = ({ device, refetch, children, canEdit }: React.PropsWit
                   <StackItem className="fctl-device-details-tab__label">{t('Labels')}</StackItem>
                 </Stack>
                 <StackItem>
-                  <EditLabelsForm device={device} onDeviceUpdate={refetch} canEdit={canEdit} />
+                  {canEdit ? (
```
celdrake, line 71: Nothing would be shown if the device is non-editable and there were no labels.

Anyway, I think it's best to split the component as `EditLabelsForm` is always meant to be used for editing the labels.

celdrake:
Closed in favor of https://github.com/flightctl/flightctl-ui/pull/208

PR #187, opened 2025-01-08
EDM-1006 Force API to produce only exact label matches

celdrake:
With the latest changes in the API for filtering, the Typeahead for searching for labels was not behaving correctly.

When typing "test" (without the equal sign), the API falls back to partial matching on label keys. To prevent the inconsistencies with having only exact match for values and partial match for keys, we change the query to only do exact matching of labels.

## Summary by CodeRabbit

- **New Features**
	- Enhanced label matching logic to support more precise API queries
	- Improved label processing for device and fleet label selections

- **Refactor**
	- Updated label utility functions to provide more exact matching capabilities
	- Removed previous label filtering method

- **Bug Fixes**
	- Improved handling of label matches to extract more accurate results

### DIFF
--- a/libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
+++ b/libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
@@ -137,10 +137,13 @@ const LabelFleetSelector = ({ selectedFleetNames, selectedLabels, onSelect, plac
   };
 
   const fetchTextMatches = async (val: string) => {
-    // Matches for labels are always exact matches
-    const labelMatches = get<DeviceList>(`devices?labelSelector=${val}&limit=1&sortBy=metadata.name&sortOrder=Asc`);
+    // By default (without the "equal" sign), the API returns a partial match, but only on the label keys
+    // To prevent this confusing behaviour, we query for exact matches in all cases (for both keys and values)
+    const exactLabelMatchParam = val.includes('=') ? val : `${val}=`;
+    const labelMatches = get<DeviceList>(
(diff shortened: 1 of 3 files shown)


libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
```diff
@@ -137,10 +137,13 @@ const LabelFleetSelector = ({ selectedFleetNames, selectedLabels, onSelect, plac
   };
 
   const fetchTextMatches = async (val: string) => {
-    // Matches for labels are always exact matches
-    const labelMatches = get<DeviceList>(`devices?labelSelector=${val}&limit=1&sortBy=metadata.name&sortOrder=Asc`);
+    // By default (without the "equal" sign), the API returns a partial match, but only on the label keys
+    // To prevent this confusing behaviour, we query for exact matches in all cases (for both keys and values)
+    const exactLabelMatchParam = val.includes('=') ? val : `${val}=`;
+    const labelMatches = get<DeviceList>(
+      `devices?labelSelector=${exactLabelMatchParam}&sortBy=metadata.name&sortOrder=Asc`,
```
rawagner, line 144: not for this PR, but shoudn't we build these `*Matches` strings using the `URLSearchParams` ?

rawagner (approved):

libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
celdrake: Yes, that's a good point.

PR #193, opened 2025-01-13
Update package.json dependencies for Ansible

celdrake:
@rawagner I was doing the same!
But also I made sure to remove all entries related to Ansible so they would all get updated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,52 +47,6 @@
         "webpack-cli": "^5.1.4"
       }
     },
-    "apps/ansible": {
-      "name": "@flightctl/ansible",
-      "version": "0.0.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -287,7 +241,6 @@
       }
     },
     "libs/ansible": {
-      "name": "@flightctl/ansible",
```
rawagner, line 290: why was this removed ? We still have libs/ansible and it has name `@flightctl/ansible`

celdrake: It's how it got recreated, I also though it was weird.
I can revert manually, but we may need to take a look.

rawagner (approved):

PR #206, opened 2025-01-31
NO-ISSUE: Enable terminal tab in ansible lib

rawagner:
## Summary by CodeRabbit

- **New Features**
	- Enhanced WebSocket endpoint generation to support device-specific connections.
	- Improved WebSocket URL construction with dynamic device ID handling.

- **Refactor**
	- Updated `useFetch` hooks across multiple applications to accept device ID.
	- Modified `useWebSocket` hook to work with device-specific endpoints.
	- Converted `useFctlNavigation` to use `React.useMemo` for performance optimization.

- **Bug Fixes**
	- Added fallback for handling undefined device names in WebSocket connections.

### DIFF
--- a/libs/ansible/src/hooks/useFetch.ts
+++ b/libs/ansible/src/hooks/useFetch.ts
@@ -46,11 +46,11 @@ export const useFetch = (getCookie: (name: string) => string | undefined, servic
   );
 
   const getWsEndpoint = React.useCallback(
-    () => ({
-      wsEndpoint: '',
+    (deviceId: string) => ({
+      wsEndpoint: `${serviceUrl}/ws/v1/devices/${deviceId}/console`,
       protocols: [],
     }),
-    [],
+    [serviceUrl],
   );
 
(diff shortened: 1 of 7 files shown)


libs/ansible/src/hooks/useFetch.ts
```diff
@@ -46,8 +46,8 @@ export const useFetch = (getCookie: (name: string) => string | undefined, servic
   );
 
   const getWsEndpoint = React.useCallback(
-    () => ({
-      wsEndpoint: '',
+    (deviceId: string) => ({
+      wsEndpoint: `/api/fctl/ws/v1/devices/${deviceId}/console`,
```
rawagner: in AAP, we do not have our proxy that provides the `/api/terminal/{id}` route (comparing to Standalone/ACM envs). We need to use the `/api/fctl/ws/v1/devices/${id}/console` route

celdrake (approved):

PR #210, opened 2025-02-10
NO-ISSUE: Fix Resource Syncs alerts

rawagner:
we switched pending with error-ed RS.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resource synchronization alerts to ensure that notifications correctly distinguish between pending and error states for clearer user feedback.


celdrake (approved):
Good catch!

PR #231, opened 2025-02-24
EDM-843: Searching devices by labels now with partial matching

celdrake:
Uses the new endpoint `GET /labels` that allows us to search for device labels including using partial matching.
(Requires https://github.com/flightctl/flightctl/pull/939)

Additionally, the search results now highlight the part of the label / fleet name which match the search text (solves EDM-533)
label-partial-matching

## Summary by CodeRabbit

- **New Features**
  - Enhanced device filtering experience by highlighting matching text in labels and fleet names.
  - Improved search functionality with support for partial label matching, offering more flexible and intuitive search results.
  - Introduced a new option for empty fleet searches to provide clearer, more predictable outcomes.
  - Added a method for constructing API-compatible strings for partial label matches.
  - Introduced a new querying function for devices based on partial label matches.

### DIFF
--- a/libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
+++ b/libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
@@ -17,7 +17,7 @@ import {
 import { TimesIcon } from '@patternfly/react-icons/dist/js/icons/times-icon';
 import { SearchIcon } from '@patternfly/react-icons/dist/js/icons/search-icon';
 
-import { DeviceList, Fleet, FleetList } from '@flightctl/types';
+import { DeviceLabelList, Fleet, FleetList } from '@flightctl/types';
 import { FlightCtlLabel } from '../../../types/extraTypes';
 import { isPromiseFulfilled } from '../../../types/typeUtils';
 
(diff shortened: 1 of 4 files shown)


libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
```diff
@@ -41,15 +41,26 @@ type LabelFleetSelectorProps = {
...
   const { t } = useTranslation();
+  const searchHighlighter = React.useCallback(
+    (text: string) => {
+      return text
+        .split(new RegExp(`(${filterText})`, 'g'))
+        .filter(Boolean)
+        .map((part, index) => (part.trim() === filterText ? <strong key={`part-${index}`}>{part}</strong> : part));
+    },
+    [filterText],
+  );
```
rawagner: Im not sure what the useCallback gives us here ? If we are after some performance optimization, IMO it would be more beneficial to not create a new regex in loop.

```suggestion
  const regexp = React.useMemo(() => new RegExp(`(${filterText})`, 'g'), [filterText]);
  const searchHighlighter = (text: string) => {
    return text
      .split(regexp)
      .filter(Boolean)
      .map((part, index) => (part.trim() === filterText ? <strong key={`part-${index}`}>{part}</strong> : part));
  };
```

celdrake: Yes, that was the idea. I agree with your suggestion, thanks

libs/ui-components/src/components/Device/DevicesPage/DeviceToolbarFilters.tsx
```diff
@@ -41,15 +41,28 @@
...
+
+  const searchHighlighter = React.useCallback(
+    (text: string) => {
+      return text
+        .split(regexp)
+        .filter(Boolean)
+        .map((part, index) => (part.trim() === filterText ? <strong key={`part-${index}`}>{part}</strong> : part));
+    },
+    [filterText],
+  );
```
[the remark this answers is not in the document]
celdrake: True, there's also a linter that caught this. Should be fixed now.

rawagner (approved):

PR #234, opened 2025-03-03
EDM-634: Make the device alias editable from the details page title

celdrake:
The device alias can now by edited from the title in the Device details page.
It cannot be edited via the labels anymore.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new device alias editor, enabling users to modify device aliases directly.

- **Enhancements**
  - Updated device details display with clearer breadcrumb titles and dynamic editing options.
  - Improved form validations and error messaging for label management.
  - Added enhanced styling support to ensure a polished look.

- **Internationalization**
  - Expanded localization strings for device aliasing, offering more intuitive instructions and error feedback.
  - Added new translation keys related to device aliasing for better user guidance.


amalykhi (approved):

libs/ui-components/src/components/Device/DeviceDetails/DeviceDetailsPage.tsx
```diff
@@ -70,7 +71,14 @@ const DeviceDetailsPage = ({ children, hideTerminal }: DeviceDetailsPageProps) =
       loading={loading}
       error={error}
       id={deviceId}
-      title={deviceAlias}
+      breadcrumbTitle={deviceAlias}
+      title={
+        canEdit ? (
+          <DeviceAliasEdit key={deviceAlias} deviceId={deviceId} alias={deviceAlias} onAliasEdited={refetch} />
```
rawagner: I think that the `key` is unnecessary ?
```suggestion
          <DeviceAliasEdit deviceId={deviceId} alias={deviceAlias} onAliasEdited={refetch} />
```

libs/ui-components/src/components/Device/DeviceDetails/DeviceAliasEdit.tsx
```diff
@@ -0,0 +1,171 @@
...
+      if (alias === originalAlias) {
+        toggleIsEditing();
+        return;
+      }
+      const req: PatchRequest = alias
+        ? [
+            {
+              path: '/metadata/labels/alias',
+              op: 'replace',
+              value: alias,
```
rawagner: I think this will fail if device's labels (or alias label) is undefined

celdrake: I wanted to prevent having to pass the existing labels, and in my testing, labels were always initialized, but I imagine there could be ways to have the labels removed and then it would fail as you mention.

I'll take a look at how to improve this.

libs/ui-components/src/components/Device/DeviceDetails/DeviceDetailsPage.tsx
celdrake: Unfortunately, this was the only way I found to prevent a small but annoying bug.

With it, every time the pencil icon is clicked, the input field has the current alias value, and the cursor starts at the end. Only when there is no alias, we show "Untitled" as the placeholder.

Without it, the first time users click on the pencil, the input field shows the current value but only as the placeholder. Meaning if users want to append something to the current alias, they need to re-type it from the beginning.  Subsequent edits behave as expected.

libs/ui-components/src/components/Device/DeviceDetails/DeviceAliasEdit.tsx
celdrake: I couldn't reproduce it, as I only saw that labels are cleared when devices are decommissioned (but those cannot be edited further).
But to be safe, I passed an extra prop to determine when the operation should be `replace` or `add`

celdrake: Last commit is amended. Now the PATCH path would be `/metadata.labels` when no labels were previously defined.

rawagner (approved):

PR #247, opened 2025-03-25
EDM-1316: Filter out duplicate ws headers

rawagner:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced WebSocket connection handling by filtering out unnecessary HTTP headers, leading to more consistent and reliable connection behavior.

### DIFF
--- a/proxy/bridge/terminal.go
+++ b/proxy/bridge/terminal.go
@@ -4,6 +4,8 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"net/textproto"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -17,6 +19,13 @@ import (
 var (
 	websocketPingInterval = 30 * time.Second
 	websocketTimeout      = 30 * time.Second
+	websocketHeaders      = []string{
+		textproto.CanonicalMIMEHeaderKey("Connection"),
+		textproto.CanonicalMIMEHeaderKey("Sec-Websocket-Extensions"),
+		textproto.CanonicalMIMEHeaderKey("Sec-Websocket-Key"),
+		textproto.CanonicalMIMEHeaderKey("Sec-Websocket-Version"),
+		textproto.CanonicalMIMEHeaderKey("Upgrade"),
+	}
 )
 
 type TerminalBridge struct {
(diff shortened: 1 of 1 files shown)


proxy/bridge/terminal.go
```diff
@@ -77,7 +85,14 @@ func (t TerminalBridge) HandleTerminal(w http.ResponseWriter, r *http.Request) {
 		TLSClientConfig: t.TlsConfig,
 	}
 
-	backend, resp, err := dialer.Dial(consoleUrl, r.Header)
+	headers := http.Header{}
+	for key := range r.Header {
+		if !slices.Contains(websocketHeaders, key) {
```
celdrake: I don't fully understand how this works.
I see you're skipping the headers defined in `websocketHeaders` and only adding those to the new `headers` object.

Does it mean that the other headers (the ones defined in `websocketHeaders`) are added by another mechanism, and that's why there were being duplicated?

proxy/bridge/terminal.go
```diff
@@ -17,6 +18,13 @@ import (
 var (
 	websocketPingInterval = 30 * time.Second
 	websocketTimeout      = 30 * time.Second
+	websocketHeaders      = []string{
+		"Connection",
```
celdrake: According to the nit-pick comment from CodeRabbit, http headers are case insensitive so it may be best to perform the comparison with ignore casing.

proxy/bridge/terminal.go
rawagner: Yes, you are correct. Since this is a proxy we have two ws connections UI <-> Proxy and Proxy <-> Backend.

When opening Proxy <-> Backend connection, the duplicate headers are added during [dialer.Dial](https://github.com/flightctl/flightctl-ui/blob/main/proxy/bridge/terminal.go#L80). You can see the implementation and the check [here](https://github.com/gorilla/websocket/blob/main/client.go#L226-L245)

celdrake: Thanks for the explanation, that makes sense!

celdrake (approved):

proxy/bridge/terminal.go
rawagner: I've updated the comparison

PR #266, opened 2025-03-31
EDM-1295: CLI Downloads page for Standalone UI

celdrake:
For Standalone UI: Adds a "help icon"  to the main toolbar, with an option to get the CLI binaries for different platforms and architectures.

Handles the case where the cliArtifacts are disabled (which happens when `FLIGHTCTL_CLI_ARTIFACTS_SERVER` is not passed as an environment variable).

Requires https://github.com/flightctl/flightctl/pull/1082 for having the downloads in the standalone deployment.

cli-downloads-now

## Summary by CodeRabbit

- **New Features**
	- Added a help dropdown in the toolbar with quick access to a new Command Line Tools page.
	- Introduced a dedicated Command Line Tools page that allows users to view and download CLI artifacts.
	- Updated user interface translations to incorporate text for command line tools and related messages.

- **Refactor**
	- Streamlined application context by replacing metrics tracking with a settings configuration.
	- Enhanced development configurations to include a new endpoint for CLI artifact retrieval.


proxy/app.go
```diff
@@ -39,6 +40,13 @@ func main() {
 	apiRouter.Handle("/flightctl/{forward:.*}", bridge.NewFlightCtlHandler(tlsConfig))
 	apiRouter.Handle("/metrics/{forward:.*}", bridge.NewMetricsHandler())
 
+	_, cliArtifactsEnabled := os.LookupEnv("FLIGHTCTL_CLI_ARTIFACTS_SERVER")
```
celdrake, line 43: When `FLIGHTCTL_CLI_ARTIFACTS_SERVER` is not defined, the Proxy will return 418  which is interpreted by the `CommandLineToolsPage` as the signal that the functionality is disabled.

libs/ansible/src/const.ts
```diff
@@ -17,4 +17,5 @@ export const appRoutes = {
   [ROUTE.RESOURCE_SYNC_DETAILS]: '/edge/resourcesyncs',
   [ROUTE.ENROLLMENT_REQUESTS]: '/edge/enrollmentrequests',
   [ROUTE.ENROLLMENT_REQUEST_DETAILS]: '/edge/enrollmentrequests',
+  [ROUTE.COMMAND_LINE_TOOLS]: '/', // TODO - TBD where to show the CLI downloads?
```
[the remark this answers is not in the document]
celdrake: This is intentional, as there should be an independent page for downloading the CLI artifacts only in the Standalone application.

Since `appRoutes` is defined as a `Record<ROUTE, string>`, we need to add an entry for the new route in `appRoutes` of each application instance, but the route itself does not exist within those applications.

libs/ui-components/src/components/Masthead/CommandLineToolsPage.tsx
```diff
@@ -0,0 +1,179 @@
...
+            <Button
+              component="a"
+              variant="link"
+              isInline
+              href={getArtifactUrl(artifactsResponse?.baseUrl || '', cliArtifact)}
+              target="_blank"
+              icon={<ExternalLinkAltIcon />}
+              iconPosition="end"
+            >
+              {t('Download flightctl CLI for {{ os }} for {{ arch }}', { os: cliArtifact.os, arch: cliArtifact.arch })}
```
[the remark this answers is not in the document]
celdrake: The CLI tool name is always "flightctl", whether it's upstream or downstream.

proxy/app.go
celdrake: Changed to statusCode=501 (Uninplemented) as Coderabbit flagged 418 statusCode (I'm a teapot) as unconventional.

apps/ocp-plugin/src/components/AppContext/AppContext.tsx
```diff
@@ -42,15 +41,18 @@ const appRoutes = {
...
 
 export const useValuesAppContext = (): AppContextProps => {
   const fetch = useFetch();
-  const metrics = useMetrics();
   const userInfo = useSelector(getUser);
 
   return {
     appType: FlightCtlApp.OCP,
+    settings: {
+      isRHEM: true,
```
rawagner: this should not be hardcoded, you can still run ocp-plugin as flightctl UI.

Do we even need this here ? OCP plugin never shows its branding anywhere

apps/standalone/src/app/hooks/useCliArtifacts.ts
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { fetchCliArtifacts } from '../utils/apiCalls';
+
+export const useCliArtifacts = () => {
+  const getCliArtifacts = React.useCallback(
+    async <R>(abortSignal?: AbortSignal): Promise<R> => fetchCliArtifacts<R>(abortSignal),
```
rawagner: Why generic `R` ? We know the format of returned value

apps/ocp-plugin/src/components/AppContext/AppContext.tsx
celdrake: We'd need to make the `isRHEM` prop optional if we want to prevent having to hardcode it.
Do you think it's better to do so ?

apps/standalone/src/app/hooks/useCliArtifacts.ts
celdrake: True.
I moved the types to `extraTypes` and the function signature in  `AppContext` already uses it.

apps/ocp-plugin/src/components/AppContext/AppContext.tsx
rawagner: yes, lets make it optional (and not set it for OCP)

apps/standalone/src/app/hooks/useCliArtifacts.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { fetchCliArtifacts } from '../utils/apiCalls';
+
+import { CliArtifactsResponse } from '@flightctl/ui-components/src/types/extraTypes';
+
+export const useCliArtifacts = () => {
```
rawagner: IMO we dont need this hook, lets remove it completely

apps/standalone/src/app/hooks/useStandaloneAppContext.ts
```diff
@@ -39,12 +39,15 @@ const standaloneAppContext: Omit<AppContextProps, 'fetch' | 'metrics'> = {
...
 
   return {
     ...standaloneAppContext,
+    settings: {
+      isRHEM: window.isRHEM || false,
+    },
     user: username,
     fetch,
-    metrics,
+    getCliArtifacts,
```
rawagner: no need to get the function via hook, lets just import it and reference it here
```suggestion
    getCliArtifacts: fetchCliArtifacts,
```

apps/standalone/src/app/utils/apiCalls.ts
```diff
@@ -64,6 +66,18 @@ export const fetchMetrics = async <R>(metricQuery: string, abortSignal?: AbortSi
   }
 };
 
+export const fetchCliArtifacts = async <R>(abortSignal?: AbortSignal): Promise<R> => {
```
rawagner:
```suggestion
export const fetchCliArtifacts = async (abortSignal?: AbortSignal): Promise<CliArtifactsResponse> => {
```

apps/standalone/src/app/hooks/useCliArtifacts.ts
celdrake: Done, as well as the other suggested changes.

apps/standalone/src/app/utils/apiCalls.ts
celdrake: I missed this left over, fixed too.

rawagner (approved):

PR #267, opened 2025-04-01
EDM-1345: Enrollment requests filter text done via API

celdrake:
The filtering of Enrollment Requests by search text was done client-side, so it could only find ERs if they happened to be in the first page of results. Now the filtering is done via the API.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined enrollment request handling to directly reflect current pending enrollments.
	- Enhanced search functionality to filter enrollment requests based on user input.
	- Updated bulk action modals to consistently operate on the most recent enrollment data.

### DIFF
--- a/libs/ui-components/src/components/EnrollmentRequest/EnrollmentRequestList.tsx
+++ b/libs/ui-components/src/components/EnrollmentRequest/EnrollmentRequestList.tsx
@@ -50,7 +50,6 @@ const EnrollmentRequestList = ({ refetchDevices, isStandalone }: EnrollmentReque
   const [canDelete] = useAccessReview(RESOURCE.ENROLLMENT_REQUEST, VERB.DELETE);
   const { remove } = useFetch();
   const [search, setSearch] = React.useState<string>('');
-  const [isLastUnfilteredListEmpty, setIsLastUnfilteredListEmpty] = React.useState<boolean>(false);
 
   const enrollmentColumns = React.useMemo(() => getEnrollmentColumns(t), [t]);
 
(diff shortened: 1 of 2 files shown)


celdrake:
There's a bug when the typed search text the list is empty, as the entire section disappears. Will review.
Update: Fixed in the second commit.

libs/ui-components/src/components/EnrollmentRequest/EnrollmentRequestList.tsx
```diff
@@ -79,8 +76,14 @@ const EnrollmentRequestList = ({ refetchDevices, isStandalone }: EnrollmentReque
...
-  // In non-standalone mode, hide the entire component when there are no pending enrollments
-  if (!isStandalone && pendingEnrollments.length === 0) {
+  React.useEffect(() => {
+    if (!search && !isLoading) {
+      setIsLastUnfilteredListEmpty(itemCount === 0);
+    }
+  }, [itemCount, search, isLoading]);
+
+  // In non-standalone mode, hide the entire component when the search result is empty (and not due to filtering)
+  if (!isStandalone && isLastUnfilteredListEmpty) {
```
rawagner: Why are you storing it in state ? Seems like we can just calculate it render loop
```suggestion
  const isLastUnfilteredListEmpty = !search && !isLoading && itemCount === 0

  // In non-standalone mode, hide the entire component when the search result is empty (and not due to filtering)
  if (!isStandalone && isLastUnfilteredListEmpty) {
```

celdrake: You are right, I'll use your patch.

rawagner (approved):

PR #290, opened 2025-05-07
EDM-1227: Overwrite OCP style for labels in Decommission device switch

celdrake:
Overwrites the CSS style created from the OCP console's CSS [here](https://github.com/openshift/console/blob/main/frontend/public/style/ancillary/_bootstrap-residual.scss#L16) which effectively styles all labels with a bold font.

loaded-style

## Summary by CodeRabbit

- **Style**
  - Improved the appearance of switch labels in device tables for better visual consistency.
  - Adjusted font weight of switch labels to match the overall design.

- **Chores**
  - Updated CSS handling to ensure consistent styling overrides within the application.

### DIFF
--- a/apps/ocp-plugin/src/components/AppContext/AppContext.css
+++ b/apps/ocp-plugin/src/components/AppContext/AppContext.css
@@ -1,0 +1,7 @@
+/* 
+  Overwrites style set by OCP's console global CSS
+  Classname is not passed to the root "label" element, we must target the PF switch first
+ */
+.pf-v5-c-switch .fctl-switch__label {
+  font-weight: normal;
+}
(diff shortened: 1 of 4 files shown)


apps/standalone/src/app/app.css
```diff
@@ -4,3 +4,10 @@ body,
   height: 100%;
 }
 
+/* 
+  Overwrites style set by OCP's console global CSS
+  Classname is not passed to the root "label" element, we must target the PF switch first
+ */
```
rawagner: shouldnt this change be scoped for ocp plugin only ? how can the standalone app.css affect plugin ?

celdrake: You are right, I was simulating the fix but I was doing so incorrectly.
I'll make sure to add the styles to the OCP plugin scope, thanks!

celdrake: I couldn't figure out a way to properly test this. Using an actual cluster and enabling the ACM operator, the operator modifies the image I'm trying to set in the UI deployment.

Theoretically, I think either loading a base CSS file from a "main" file such as ocp-plugin's AppContext, or wrapping all routes in a common component should work.

Any ideas?

rawagner: using `AppContext` sounds good to me

celdrake: Cool, updating moving the CSS there.

rawagner (approved):

PR #297, opened 2025-05-21
EDM-1580 extend Events system MVP

celdrake:
Extends the Event system MVP by adding a virtualized list for better performance.
Needs to be evaluated when the Backend part is complete.


celdrake:
Will need revisiting since the "VirtualizedTable" approach will soon be deprecated at the OCP console.
They suggest using the newish "Data view" from Patternfly but that's more geared towards tables than lists.

PR #300, opened 2025-05-26
Do not discard changes in Updates page

rawagner:
This PR simplifies the behavior of Update step. We keep the user changes when the checkboxes are being de-selected.

## Summary by CodeRabbit

- **Removals**
  - Deprecated localization strings related to switching to basic update configurations and removing rollout, disruption, and update policies.
  - Removed the confirmation modal component for updating policies.
- **Improvements**
  - Simplified UI logic by removing complex toggle states and modal interactions for update configuration modes.
  - Updated components to directly bind checkboxes to configuration flags, streamlining user interactions.
  - Reversed display logic in review components to show batch counts for advanced policies.
- **Validation & Data Handling**
  - Implemented conditional validation schemas that depend on the `useBasicUpdateConfig` flag.
  - Updated form initial values and patch functions to align with the new configuration logic.
- **Existing Features**
  - The toggle for basic and advanced update configurations remains, now with simplified control flow.


libs/ui-components/src/components/Fleet/CreateFleet/steps/UpdatePolicyStep.tsx
```diff
@@ -26,124 +21,31 @@ const UpdatePolicyStep = ({ isReadOnly }: { isReadOnly: boolean }) => {
...
-      }
-    } else {
-      setAlertSwitchToBasic(setting);
-    }
-  };
-
-  const onChangePolicyType = async (toAdvanced: boolean) => {
-    await setFieldValue('rolloutPolicy', {
-      isAdvanced: toAdvanced,
-      batches: toAdvanced ? [getEmptyInitializedBatch()] : [],
```
celdrake, line 64: There's an issue now, where it's possible to get to the Review page with a rollout policy that has no batches. Before, the empty batch would be initialized by default.

When attempting to PATCH the fleet, the backend returns an error 400 because of that.

celdrake (commented):
I like the new flow, it's definitely simpler!
Did you check with Nir that it's OK not to have any kind of message? I understand it may not be necessary anymore since the action is reversable.

libs/ui-components/src/components/Fleet/CreateFleet/steps/UpdatePolicyStep.tsx
rawagner: Updated, I am adding the init batch to init values if there isn't any.

rawagner:
> I like the new flow, it's definitely simpler! Did you check with Nir that it's OK not to have any kind of message? I understand it may not be necessary anymore since the action is reversable.

Nir is OK with the change. He will also update the figma to make sure it is aligned with the implementation.

libs/ui-components/src/components/form/validations.ts
```diff
@@ -457,7 +457,7 @@ export const validFleetRolloutPolicySchema = (t: TFunction) => {
       }
 
       const errors = rolloutPolicy.batches.reduce((errors, batch, batchIndex) => {
-        const hasError = batch.limit === undefined && Object.keys(batch.selector || {}).length === 0;
+        const hasError = batch.limit === undefined && batch.selector.length === 0;
         if (hasError) {
```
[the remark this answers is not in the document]
rawagner: TS type defines this as an array that must be defined. `undefined` value is not allowed

celdrake (approved):

PR #312, opened 2025-06-06
Fix dependency install package lock

celdrake:
Fixes several issues that appeared when performing a clean install to re-generate `package-lock.json`

- Webpack version would become incompatible with the projects' webpack.config.ts files
- Eslint becomes updated and starts flagging certain regular expressions that use named arguments
- i18next-parser becomes updated and updates "cheerio" to 1.0.0 which introduced a breaking change and the app won't compile
- Updates to "react-router" libraries prevent `npm run build:ocp` to complete. 

Note: I need to verify that this solution works well in the deployed environments, I'll remove the draft when that's done.

### DIFF
--- a/apps/ocp-plugin/package.json
+++ b/apps/ocp-plugin/package.json
@@ -80,7 +80,6 @@
     "react-dom": "17.0.1",
     "react-i18next": "^11.7.3",
     "react-redux": "7.2.2",
-    "react-router-dom": "^5.3.x",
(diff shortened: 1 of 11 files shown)


celdrake:
The changes related to the `react-router` are not working when the UI is deployed as the OCP plugin.
Need to re-evaluate.

celdrake:
Closing since few dependencies have changed and we'd need to revisit this.

PR #319, opened 2025-06-25
EDM-1682: Allow for templated SystemD services

SiddarthR56:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for systemd unit patterns, allowing a wider range of valid characters and formats, including support for templated and glob-style unit names.

### DIFF
--- a/libs/ui-components/src/components/form/validations.ts
+++ b/libs/ui-components/src/components/form/validations.ts
@@ -29,7 +29,8 @@ import {
 import { labelToString } from '../../utils/labels';
 import { UpdateScheduleMode } from '../../utils/time';
 
-const SYSTEMD_PATTERNS_REGEXP = /^[a-z][a-z0-9-_.]*$/;
+const SYSTEMD_PATTERNS_REGEXP =
+  /^[0-9a-zA-Z:\-_.\\\[\]!\-\*\?]+(@[0-9a-zA-Z:\-_.\\\[\]!\-\*\?]+)?(\.[a-zA-Z\[\]!\-\*\?]+)?$/;
 const SYSTEMD_UNITS_MAX_PATTERNS = 256;
 
 // Accepts uppercase characters, and "underscore" symbols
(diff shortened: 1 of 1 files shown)


libs/ui-components/src/components/form/validations.ts
```diff
@@ -29,7 +29,7 @@ import {
 import { labelToString } from '../../utils/labels';
 import { UpdateScheduleMode } from '../../utils/time';
 
-const SYSTEMD_PATTERNS_REGEXP = /^[a-z][a-z0-9-_.]*$/;
+const SYSTEMD_PATTERNS_REGEXP = /^[0-9a-zA-Z:-_.\\\[\]!-*?]+(@[0-9a-zA-Z:-_.\\\[\]!-*?]+)?(\.[a-zA-Z\[\]!-*?]+)?$/;
```
celdrake: Thanks for your PR, @SiddarthR56 !

However, it looks like this RegExp is not the same as in the Backend PR, as the example "api-server@.service" is accepted on the backend, but not on the UI.

libs/ui-components/src/components/form/validations.ts
```diff
@@ -653,8 +653,8 @@ export const systemdUnitListValidationSchema = (t: TFunction) =>
       }),
     )
     .test('invalid patterns', (systemdUnits: SystemdUnitFormValue[] | undefined, testContext) => {
-      // TODO analyze https://github.com/systemd/systemd/blob/9cebda59e818cdb89dc1e53ab5bb51b91b3dc3ff/src/basic/unit-name.c#L42
-      // and adjust the regular expression and / or the validation to accommodate for it
+      // Supports templated SystemD services with extended regex for all allowed unit file formats and glob searches
+      // Pattern supports: [0-9a-zA-Z:-_.\\\[\]!-*?]+(@[0-9a-zA-Z:-_.\\\[\]!-*?]+)?(\.[a-zA-Z\[\]!-*?]+)?
```
celdrake:
```suggestion
      // See SYSTEMD_PATTERNS_REGEXP
```
as otherwise the comment may easily become outdated

Or maybe move the comment on top of the RegExp so anyone checking it understands what type of systemd services it allows.

celdrake (changes requested):

celdrake (approved):

PR #330, opened 2025-07-30
EDM-1654: Show alerts in Overview page

celdrake:
Add the alerts card to the Overview page.

It will be only shown if:
- Alertmanager service is enabled
- The user has permissions for viewing the alerts.

Requires: 
https://github.com/flightctl/flightctl/pull/1454 and https://github.com/flightctl/flightctl/pull/1452 in the Backend.

Tested in the OCP plugin, with the backend fixes. The alert was generated by inserting a record in the DB directly, the data is invented and has some mistakes.

alerts-ocp-plugin

## Summary by CodeRabbit

* **New Features**
  * Introduced an Alerts card to the Overview page, displaying active alerts with localized messages and resource links.
  * Added empty state messaging and navigation when no alerts are present.
  * Implemented periodic and manual refresh for alerts data.
  * Alerts feature is conditionally enabled based on user access and backend availability.

* **Bug Fixes**
  * Improved error handling and user feedback for alert loading and API errors.

* **Chores**
  * Updated environment setup scripts for development, consolidating IP detection and environment variable configuration.
  * Enhanced localization with new alert-related translation strings.


apps/ocp-plugin/src/utils/apiCalls.ts
```diff
@@ -56,6 +60,35 @@ export const fetchMetrics = async <R>(metricQuery: string, abortSignal?: AbortSi
   }
 };
 
+export const fetchAlerts = async <R>(abortSignal?: AbortSignal): Promise<R> => {
```
rawagner, line 63: why is this function using generics ? The Alerts return is known, right ?

libs/ui-components/src/hooks/useAlerts.ts
```diff
@@ -0,0 +1,147 @@
...
+    inhibitedBy: string[];
+    mutedBy: string[];
+    silencedBy: string[];
+  };
+  receivers: Array<{ name: string }>;
+};
+
+const DEFAULT_REFRESH_INTERVAL = 30000; // 30 seconds
+
+export const useAlerts = (
```
rawagner, line 30: why are we not reusing `useFetchPeriodically` hook ?

rawagner (approved):
Looks good, we can address the comments in followups

libs/ui-components/src/hooks/useAlerts.ts
celdrake: I didn't reuse it initially because I was using this hook to detect when alerts were disabled (we know in the UI as it returns 501), and in that case we didn't want to keep fetch .

Since later I introduced "useAlertsEnabled", the main hook can use useFetchPeriodically. Will refactor later.

apps/ocp-plugin/src/utils/apiCalls.ts
celdrake: True, I'll fix it.

PR #346, opened 2025-09-17
EDM-1781 multi org support

celdrake:
Adds support for multiple organizations within Flight Control.

1. If organizations are disabled, the system behaves as until now.
2. If organizations are disabled, we check if the user is required to select an organization. If they only belong to one, the organization is auto-selected. Otherwise, they are required to select an organization before they can access the UI.

Technical notes:
- The initial implementation to detect if the user  has selected an organization was based on a Cookie managed by the UI Proxy. However, this didn't seem to work in the OCP plugin as the cookies were not arriving to the proxy.
- The second commit changes the implementation to be based on `localStorage` and headers. WIth this approach, there is only difference in that in OCP plugin we cannot identify when we need to clear the previous selection. Meaning we won't prompt for the user to select a new organization if they had selected one previously. 

Organizations enabled:
https://github.com/user-attachments/assets/1a4a1f01-c636-4601-b05f-dc6ed45bd92f

Organizations disabled:
https://github.com/user-attachments/assets/adb14db5-c513-4234-921b-9e6f975c61fa

Requires https://github.com/flightctl/flightctl/pull/1672

## Summary by CodeRabbit

- **New Features**
  - Multi-organization support: organization selector on first login and "Change Organization" in navigation; org-aware UI/messages and gating.
  - UI proxy endpoints: app routes API and alerts via a UI proxy and propagates organization header.

- **Refactor**
  - Consistent page layout wrapper applied across many pages; layout gated by organization guard and selection state.

- **Chores**
  - Dev toggle to enable Organizations in standalone builds; new translations for organization flows.


apps/ocp-plugin/src/components/common/WithPageLayout.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import OrganizationGuard from '@flightctl/ui-components/src/components/common/OrganizationGuard';
+
+// Restore WithPageLayoutContent when organizations are enabled for OCP plugin
```
celdrake, line 4: Commented out since last news is that there won't be support for OCP Plugin in v0.10.

In the tests I did, we only have a single organization. 
The UI behaves correctly:
 - The proxy requests have the x-flightctl-organization-id header
 - In the API logs, the "org_id" query parameter can be seen with the default organization ID.

libs/ui-components/src/components/common/OrganizationGuard.tsx
```diff
@@ -0,0 +1,129 @@
...
+        // First, check if organizations are enabled via proxy endpoint
+        const organizationsEnabledResponse = await proxyFetch('organizations-enabled', {
+          method: 'GET',
+          headers: { 'Content-Type': 'application/json' },
+          credentials: 'include',
+        });
+
+        const organizationsEnabled = organizationsEnabledResponse.status !== 501;
+        setIsOrganizationsEnabled(organizationsEnabled);
+
```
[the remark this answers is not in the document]
celdrake: The statusCode 501 is explicitly returned by the UI proxy based on the ORGANIZATIONS_ENABLED setting.
Any other statusCode means organizations are enabled.

proxy/middleware/organization.go
```diff
@@ -0,0 +1,76 @@
...
+	organizationID := r.Header.Get("X-FlightCtl-Organization-ID")
+
+	if organizationID == "" {
+		log.GetLogger().Info("DEBUG MIDDLEWARE: No organization header found")
+		return "", fmt.Errorf("no organization header found")
+	}
+
+	log.GetLogger().Infof("DEBUG MIDDLEWARE: Organization from header: '%s'", organizationID)
+	return organizationID, nil
+}
```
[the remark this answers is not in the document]
celdrake: I'll remove the logs.

proxy/middleware/organization.go
```diff
@@ -0,0 +1,72 @@
...
+// OrganizationMiddleware adds org_id query parameter to FlightCtl API requests
+// and blocks API calls when no organization is selected
+// Note: This middleware is only loaded when organizations are enabled
+func OrganizationMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Only process FlightCtl API calls that should have org_id
+		if !shouldAddOrgId(r.URL.Path) {
+			next.ServeHTTP(w, r)
+			return
+		}
```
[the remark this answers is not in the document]
celdrake: Added

proxy/middleware/organization.go
```diff
@@ -0,0 +1,64 @@
...
+
+const (
+	headerOrganizationID = "X-FlightCtl-Organization-ID"
+	queryOrganizationID  = "org_id"
+)
+
+// OrganizationMiddleware adds org_id query parameter to FlightCtl API requests
+// and blocks API calls when no organization is selected
+
+func OrganizationMiddleware(next http.Handler) http.Handler {
```
celdrake, line 16: FYI "alerts" endpoint is broken ATM.
With the following [PR](https://github.com/flightctl/flightctl/pull/1627), it should be fixed again. 

Then I'll need to make adaptations to ensure we're also passing the org_id header on the alerts endpoint.

libs/ui-components/src/components/common/OrganizationSelector.tsx
```diff
@@ -0,0 +1,205 @@
...
+  React.useEffect(() => {
+    if (defaultOrganizationId && !hasUserMadeSelection.current && !selectedOrg) {
+      setSelectedOrg(defaultOrganizationId);
+    }
+  }, [defaultOrganizationId, selectedOrg]);
+
+  const handleOrgSelection = React.useCallback((orgId: string) => {
+    hasUserMadeSelection.current = true;
+    setSelectedOrg(orgId);
+  }, []);
```
rawagner: Could this be simplified to the following..? 

```suggestion
  const [selectedOrg, setSelectedOrg] = React.useState<string>();
  const needsScroll = organizations.length > MAX_ORGANIZATIONS_FOR_SCROLL;

  // Only set default organization if user hasn't made a manual selection yet
  React.useEffect(() => {
    if (defaultOrganizationId && !selectedOrg) {
      setSelectedOrg(defaultOrganizationId);
    }
  }, [defaultOrganizationId, selectedOrg]);

  const handleOrgSelection = React.useCallback((orgId: string) => {
    setSelectedOrg(orgId);
  }, []);
```

proxy/middleware/organization.go
```diff
@@ -0,0 +1,64 @@
...
+		query := r.URL.Query()
+		query.Set(queryOrganizationID, orgID)
+		r.URL.RawQuery = query.Encode()
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func isFlightCtlAPICall(path string) bool {
+	return strings.Contains(path, "/api/v1/")
```
rawagner: Lets use `HasPrefix` here. The `Contains` is a bit loose IMO.

proxy/middleware/organization.go
```diff
@@ -0,0 +1,64 @@
...
+	excludePaths := []string{
+		// The API call to retrieve all organizations (otherwise, if called with an org_id, only that particular organization will be returned
+		"/api/v1/organizations",
+		"/login",
+		"/logout",
+		"/config",
+	}
+
+	for _, excludePath := range excludePaths {
+		if strings.Contains(path, excludePath) {
```
rawagner: I'd prefer `HasPrefix` here too.

proxy/middleware/organization.go
celdrake: Alerts endpoint was fixed. This required further adaptations in the UI.

proxy/middleware/organization.go
```diff
@@ -0,0 +1,67 @@
...
+			// No organization selected - block this API call with 428 Precondition required
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusPreconditionRequired)
+			w.Write([]byte(`{"error": "Organization selection required", "code": "ORGANIZATION_REQUIRED"}`))
+			return
+		}
+
+		query := r.URL.Query()
+
+		if isAlertsAPICall(r.URL.Path) {
```
celdrake, line 36: I made some changes to fix the alerts endpoint calling from the UI.

proxy/middleware/organization.go
celdrake: Done. ANd I made some further changes to allow "alerts" to have the header too.

proxy/app.go
```diff
@@ -21,7 +21,7 @@ func corsHandler(router *mux.Router) http.Handler {
 	return gorillaHandlers.CORS(
 		gorillaHandlers.AllowedOrigins([]string{"http://localhost:9000"}),
 		gorillaHandlers.AllowedMethods([]string{"GET", "POST", "PUT", "DELETE", "PATCH"}),
-		gorillaHandlers.AllowedHeaders([]string{"Content-Type", "Authorization"}),
+		gorillaHandlers.AllowedHeaders([]string{"Content-Type", "Authorization", "X-FlightCtl-Organization-ID"}),
```
celdrake, line 24: For some reason this didn't fail before, but now all requests were being rejected with CORS errors.
Adding the header to the allowed headers fixed it.

libs/ui-components/src/components/common/OrganizationSelector.tsx
celdrake: I actually removed the effect. I remember at some point there was a timing issue that defaultORg wasn't being received when the component mounted initially, but I couldn't see it happening anymore.

libs/ui-components/src/components/common/OrganizationSelector.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <Menu
+          activeItemId={selectedOrg}
+          selected={selectedOrg}
+          onSelect={(_ev, orgId) => handleOrgSelection(orgId as string)}
+          isScrollable={needsScroll}
+        >
+          <MenuContent menuHeight={needsScroll ? '230px' : 'auto'}>
+            <MenuList>
+              {organizations.map((org) => {
+                const orgId = org.metadata?.name as string;
```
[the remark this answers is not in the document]
celdrake: The organization "metadata.name" is its ID, so it should always have one.

libs/ansible/src/hooks/useFetch.ts
```diff
@@ -52,6 +52,13 @@ export const useFetch = (getCookie: (name: string) => string | undefined, servic
 
   const checkPermissions = React.useCallback(() => Promise.resolve(true), []);
 
+  const proxyFetch = React.useCallback(
+    async (endpoint: string, requestInit: RequestInit): Promise<Response> => {
+      return fetchUiProxy(endpoint, serviceUrl, applyHeaders, requestInit);
+    },
+    [serviceUrl, applyHeaders],
+  );
```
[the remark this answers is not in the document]
celdrake: I added a separate commit to prevent mutating the header objects across different files and functions.
Could you check if that fixes this issue everywhere?

libs/ansible/src/hooks/useFetch.ts
```diff
@@ -1,17 +1,18 @@
...
-          options.headers['X-CSRFToken'] = token;
-        } else {
-          options.headers = { 'X-CSRFToken': token };
-        }
+        const headers = new Headers(options.headers || {});
+        headers.set('X-CSRFToken', token);
+        return {
+          ...options,
+          headers,
+        };
```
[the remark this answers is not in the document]
celdrake: The ansible implementation ATM is not either finalized or being used. 
There's no point in making changes to this library unless it's something that will break when building the images etc.

rawagner (approved):

PR #356, opened 2025-10-01
Adjust new subnav bar to UX designs

celdrake:
The current design had some misalignments to the UX designs:
- Subnav should be in the same background color as the left navigation menu
- Subnav height should be the same as the first element  in the left navigation menu
- The "Change organization" dropdown item text was slightly cut

Before the changes:
subnav-before

After the changes:
subnav-fixed

## Summary by CodeRabbit

* **Refactor**
  * Replaced header toolbar with a Masthead layout; page navigation now only shows the organization selector when multiple organizations exist and no longer accepts nested children.

* **Style**
  * Removed padding from the main navigation list.
  * Aligned toolbar and subnav content to the end.
  * Added a global actions masthead region with zero padding and standardized height.
  * Applied a darker masthead/subnav background.

* **Enhancements**
  * Organization dropdown supports right-side positioning for improved usability.

### DIFF
--- a/libs/ui-components/src/components/common/PageNavigation.tsx
+++ b/libs/ui-components/src/components/common/PageNavigation.tsx
@@ -3,6 +3,8 @@ import {
   Dropdown,
   DropdownItem,
   DropdownList,
+  Masthead,
+  MastheadContent,
   MenuToggle,
   MenuToggleElement,
   PageSection,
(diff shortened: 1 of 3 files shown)


libs/ui-components/src/components/common/PageNavigation.tsx
```diff
@@ -68,21 +69,24 @@ const PageNavigation = ({ children }: React.PropsWithChildren) => {
...
-                />
-              </ToolbarItem>
-            )}
-          </ToolbarContent>
-        </Toolbar>
+        <Masthead id="global-actions-masthead">
+          <MastheadContent>
+            <Toolbar isFullHeight isStatic className="fctl-subnav_toolbar">
+              <ToolbarContent>
+                {showOrganizationSelection && (
```
celdrake, line 76: Keeping this condition despite being redundant, since it won't be once we add the new "Copy login command" option to the navigation bar.

rawagner (approved):

PR #367, opened 2025-10-10
Migrate ts-codegen to @hey-api/openapi-ts

celdrake:
The `openapi-typescript-codegen` is now deprecated, the recommended alternative is `@hey-api/openapi-ts`
I saw this when I was reviewing options for adding schema suggestions etc to the YAML editor. Seems like the "schema.gen.ts" that can be generated could be useful for this.

There are some differences in how the types are generated:
1. Now the types are all created in a single file, `types.gen.ts`, which is later renamed as `index.ts`
2. ConditionType is now being generated correctly
3. DeviceSystemInfo is generated with incorrect "additionalProperties" of type `CustomInfo`. I added a transformer that modifies the schema so it's generated correctly (as [before](https://github.com/flightctl/flightctl-ui/blob/main/libs/types/models/DeviceSystemInfo.ts#L8)).
4. RolloutDeviceSelection is generated incorrectly, it should be a pure alias to BatchSequence (as [before](https://github.com/flightctl/flightctl-ui/blob/main/libs/types/models/RolloutDeviceSelection.ts#L9)). I added a transformer for this too.
5. There is a warning related to ESlint now when running "npm run lint", but it all seems to work fine.


libs/types/scripts/openapi-typescript.mjs
```diff
@@ -0,0 +1,45 @@
...
+  input: jsonSchema,
+  output: {
+    path: './tmp-types',
+    indexFile: false,
+  },
+  plugins: [
+    {
+      name: '@hey-api/typescript',
+      enums: {
+        mode: 'typescript',
```
celdrake, line 17: By default, enums would be created as types which would break a lot of our code.
Also, by default it uses the `SCREAMING_SNAKE_CASE` casing which would have required a lot more changes.

With the previous generator, only some types were using `SCREAMING_SNAKE_CASE` (eg. Event.type, etc)

celdrake:
@coderabbitai The gist of this PR was to change the code generator for the Typescript types definitions.
The old generator produced one file per type, under /libs/types/models.
The new generator produces a single file for all types, which we rename to "index.ts"

There's no need to review the changes of the files removed at libs/types/models folder.

celdrake:
We'll review in the future if this is needed. Closing for now.

PR #375, opened 2025-11-11
Adding application type Quadlet support

liatb-rh:
Inline applications now support Compose/Quadlet in the editor, appType is preserved from the API, and emitted on save.
Vibe coding
Screenshot 2025-11-11 at 16 49 41

## Summary by CodeRabbit

* **New Features**
  * Added application format selector (Compose/Quadlet) for inline applications in the device wizard.
  * Application type preferences are now preserved when switching between application types.

### DIFF
--- a/libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationTemplates.tsx
+++ b/libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationTemplates.tsx
@@ -1,6 +1,9 @@
 import * as React from 'react';
 
-import { Button, FormGroup, FormSection, Grid, Split, SplitItem } from '@patternfly/react-core';
+import { Button } from '@patternfly/react-core/dist/esm/components/Button';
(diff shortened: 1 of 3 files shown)


libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationTemplates.tsx
```diff
@@ -45,6 +48,7 @@ const ApplicationSection = ({ index, isReadOnly }: { index: number; isReadOnly?:
       setValue(
         {
           specType: AppSpecType.INLINE,
+          appType: (app.appType || ('compose' as unknown as AppForm['appType'])),
```
celdrake, line 51: Seems like it would work without the casting
```suggestion
          appType: app.appType,
```

celdrake (commented):
We'd be missing validations since for quadlet applications we only want to allow the user to set certain types (eg. .container, .volume), while attempting to save other types (eg. .kube) would get a 400 on the API side.

celdrake:
Full implementation here https://github.com/flightctl/flightctl-ui/pull/381

PR #409, opened 2025-12-03
EDM-2730: Redirect to valid page after switching organization

celdrake:
When users switch organization while logged in and while viewing a resource details page, after the switch they don't have access to the same resource as it belongs to the previous organization.

Now we change the page to either the parent page (eg. view repository --> repository list), or Overview as a fallback.

https://github.com/user-attachments/assets/bfa6b696-689b-4765-9efb-00806b924f19

## Summary by CodeRabbit

* **Bug Fixes**
  * Organization switching now uses client-side redirects instead of a full reload, preserving user context and mapping detail/edit pages to their appropriate list/overview routes so users end up on the correct overview after switching.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/libs/ui-components/src/components/common/PageNavigation.tsx
+++ b/libs/ui-components/src/components/common/PageNavigation.tsx
@@ -21,6 +21,7 @@ import { useOrganizationGuardContext } from './OrganizationGuard';
 import OrganizationSelector from './OrganizationSelector';
 import { RESOURCE, VERB } from '../../types/rbac';
 import { usePermissionsContext } from './PermissionsContext';
+import { useAppContext } from '../../hooks/useAppContext';
 
 import './PageNavigation.css';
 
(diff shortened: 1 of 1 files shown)


libs/ui-components/src/components/common/PageNavigation.tsx
```diff
@@ -63,9 +64,30 @@ const OrganizationDropdown = ({ organizationName, onSwitchOrganization }: Organi
...
+  if (deviceEntityMatch) {
+    return `/${deviceEntityMatch[1]}/${deviceEntityMatch[2]}`;
+  }
+
+  // In the admin section, currently there's only one option, but we keep it generic for future expansion.
+  const adminMatch = pathname.match(/^\/admin\/(authproviders)\/[^/]+/);
+  if (adminMatch) {
+    const entityType = adminMatch[1];
+    return `/admin/${entityType}`;
+  }
```
rawagner: could we instead use routes defined via `AppContext.appRoutes` ?

celdrake: good idea, I'll do.

celdrake: Done

rawagner (approved):

PR #431, opened 2025-12-12
Do not show warning icon for unsupported integrity status

celdrake:
Needs confirmation from UX that we should show also the InfoIcon for the integrity status higher level.
integrity-all


celdrake:
Needs a proper assessment with UX. Closing for now.

PR #448, opened 2025-12-19
EDM-2865: Upgrade to Patternfly v6

celdrake:
Yet to be done: review if the YAML editor is aligned with newer versions.


libs/ui-components/src/components/AuthProvider/AuthProviderDetails/RoleAssigmentDetails.tsx
```diff
@@ -28,15 +28,15 @@ const RoleAssigmentDetails = ({ roleAssignment }: { roleAssignment: AuthRoleAssi
   return (
     <>
       <Label color="purple">{getAssignmentTypeLabel(roleAssignment.type, t)}</Label>
-      <LabelGroup className="pf-v5-u-mt-sm">
+      <LabelGroup className="pf-v6-u-ml-sm">
```
celdrake, line 31: it was missing horizontal spacing

libs/ui-components/src/components/form/validations.ts
```diff
@@ -596,11 +596,11 @@ export const validateCPULimit = (cpu: string | undefined): boolean => {
   if (!cpu) {
     return true;
   }
-  const trimmed = cpu.trim();
-  if (!trimmed) {
-    return true;
+
+  if (!/^[0-9]+$/.test(cpu)) {
```
jgyselov: This change results in the suggested value `0.4` being invalid. Not sure what's correct here but either the validation or the helper text needs to change.

Image

jgyselov (changes requested):
Just one potential little problem. 🙂 

As for the PF6 updates, I didn't see any issues in the code or the UI (the parts that I managed to interact with).

libs/ui-components/src/components/form/validations.ts
celdrake: Good catch, thanks!

This was so that 3433asf would be considered an invalid value, but I forgot about the fractions. I'll double check that it's enough to add the `dot` to the regular expression.

Could be part of another PR but I guess since I was testing all fields etc I just noticed that.

celdrake:
> Just one potential little problem. 🙂
> 
> As for the PF6 updates, I didn't see any issues in the code or the UI (the parts that I managed to interact with).

Great, thanks! 

The "Devices" area is a bit tricker to test because you need to either enroll a device (it can be done easily in the backend project doing `make agent-vm`, but it's a bit slow), or simulate one (but then you can't have the device change its status, etc).

Let me know if you'd want to proceed further or you're satisfied with what you covered.

libs/ui-components/src/components/form/validations.ts
jgyselov: Up to you :slightly_smiling_face:

jgyselov:
> > Just one potential little problem. 🙂
> > As for the PF6 updates, I didn't see any issues in the code or the UI (the parts that I managed to interact with).
> 
> Great, thanks!
> 
> The "Devices" area is a bit tricker to test because you need to either enroll a device (it can be done easily in the backend project doing `make agent-vm`, but it's a bit slow), or simulate one (but then you can't have the device change its status, etc).
> 
> Let me know if you'd want to proceed further or you're satisfied with what you covered.

I managed to simulate a fleet so I could see the tables, the detail pages, and the edit forms etc. It all looked fine and the changes in the code weren't complicated anyway. I think you're good here.

libs/ui-components/src/components/form/validations.ts
celdrake: Fixed now

celdrake:
@jgyselov I fixed the CPU validation issue, rebased to latest "main" (and adapted the new LoginCommand component to PF6).
IMO this should be good now, could you please take a final look?

jgyselov (approved):

jgyselov (approved):

PR #453, opened 2026-01-13
EDM-3069 patch credentials only if changed

celdrake:
Whenever the users would modify an OCI registry, the credentials were incorrectly included (as they are masked)

The same thing is likely happening for other repository types, but there's a bug in the API (EDM-3068) and we can't set credentials on those repositories to test it properly. Marking as Draft until the backend bug is fixed.

### DIFF
--- a/libs/ui-components/src/components/Repository/CreateRepository/utils.ts
+++ b/libs/ui-components/src/components/Repository/CreateRepository/utils.ts
@@ -217,12 +217,6 @@ export const getRepositoryPatches = (values: RepositoryFormValues, repository: R
         if (ociRepoSpec.skipServerVerification !== undefined) {
           patches.push({ op: 'remove', path: '/spec/skipServerVerification' });
         }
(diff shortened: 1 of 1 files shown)


libs/ui-components/src/components/Repository/CreateRepository/utils.ts
```diff
@@ -217,12 +216,6 @@ export const getRepositoryPatches = (values: RepositoryFormValues, repository: R
         if (ociRepoSpec.skipServerVerification !== undefined) {
           patches.push({ op: 'remove', path: '/spec/skipServerVerification' });
         }
-        if (ociRepoSpec.scheme) {
```
celdrake, line 220: `scheme` and `accessMode` are not part of the "additionalSettings" field, they are mandatory though they have defauts. S this would have been resetting to the defaults incorrectly.

celdrake:
Has been incorporated in https://github.com/flightctl/flightctl-ui/pull/459

PR #487, opened 2026-02-09
EDM-3315: Fix wrong request to download ImageBuilder API after redirect

celdrake:
The flow for downloading an image export image was the following:
1. Request to ImageBuilder API "download" endpoint with "manual" redirect to prevent CORS issues.
2. The response included the "Location" header, but the UI was not able to read it.
3. As a result, it triggered a second request to the same "download" endpoint, but this time it didn't include the `X-FlightCtl-Organization-ID` header, triggering error 428 in the UI proxy.

To fix this, now we use the UI proxy to read the Location header and return the response as a statusCode 200 with the redirect URL in the response body.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined download and redirect handling for image build exports to simplify control flow while preserving user-visible behavior.
  * Centralized detection of redirect responses so the UI reliably receives redirect links or blob downloads.

* **Bug Fixes**
  * Improved robustness for edge cases (invalid responses, opaque redirects, missing filenames) to reduce download errors.


rawagner (approved):

celdrake:
@rawagner Had to fix conflicts. Could you please re-approve?

rawagner (approved):

PR #496, opened 2026-02-10
EDM-3291: ignore empty enums in openapi

sk-ilya:


libs/types/models/SystemdEnableStateType.ts
```diff
@@ -20,5 +20,4 @@ export enum SystemdEnableStateType {
   SystemdEnableStateTransient = 'transient',
   SystemdEnableStateBad = 'bad',
   SystemdEnableStateUnknown = 'unknown',
-  SystemdEnableStateEmpty = '',
```
celdrake, line 23: As discussed, the problem is that in this case `SysetmEnableStateEmpty` is a valid case that we must preserve.

PR #497, opened 2026-02-11
EDM-3307 helm apps validations part2

celdrake:
The previous PR had some gaps relative to helm applications validations:
- Helm apps "name" is optional
- Images may not start with "dot", "slash" or "undescore". This is the same for all OCI references (application images, volume images, etc).

## Summary by CodeRabbit

* **New Features**
  * Stricter OCI image name validation with clearer error messages for invalid characters and disallowed leading characters.

* **Bug Fixes**
  * Application name field is no longer marked required in the Helm application form.
  * Empty or unset application names are omitted from API payloads instead of sent as empty values.

* **Documentation**
  * Updated user-facing validation messages/localization for image errors.


libs/ui-components/src/components/form/validations.ts
```diff
@@ -678,12 +687,11 @@ const mountPathSchema = (t: TFunction, isRequired: boolean = true) => {
   return isRequired ? schema.required(t('Mount path is required for this volume type')) : schema;
 };
 
-// Volume validation schemas
 const singleContainerVolumesSchema = (t: TFunction) =>
   Yup.array().of(
     Yup.object().shape({
       name: volumeNameSchema(t),
-      imageRef: optionalImageRefSchema(t),
+      imageRef: ociImageSchema(t),
```
celdrake, line 695: There's a bug here, this is still optional. Will fix.

celdrake: Fixed @coderabbitai can you verify?

celdrake: so all cases where image was optional are still optional, and same for required fields

rawagner (approved):

PR #516, opened 2026-02-17
Update react-router

celdrake:
Seems to fail as the OCP plugin, under review.


celdrake:
Solved differently in https://github.com/flightctl/flightctl-ui/pull/528, with less drastic changes.

PR #533, opened 2026-02-19
FCTL catalog

rawagner:
## Summary by CodeRabbit

* **New Features**
  * Added Software Catalog navigation menu item for managing applications and system images
  * Introduced Catalog page displaying available software items with search and filtering
  * Added multi-step install wizard for deploying OS images and applications to devices and fleets
  * Created edit wizards for updating installed software with version selection and change review
  * Added installed software view showing currently deployed catalog items on devices and fleets
  * Implemented update visualization showing available version upgrades with dependency tracking

* **Enhancements**
  * Extended device and fleet details pages with Catalog tabs for software management


apps/standalone/src/app/routes.tsx
```diff
@@ -150,6 +155,31 @@ const getAppRoutes = (t: TFunction): ExtendedRouteObject[] => [
       </TitledRoute>
     ),
   },
+  {
+    path: '/catalog',
```
celdrake: According to Figma, the `Catalog` entry should be after "Image builds"

libs/types/models/ConditionType.ts
```diff
@@ -7,17 +7,16 @@
  */
 export enum ConditionType {
```
celdrake, line 8: The file needs to be reverted and checked manually, there's likely no new ConditionTypes

libs/types/scripts/openapi-typescript.js
```diff
@@ -69,6 +74,26 @@ async function generateTypes(mode) {
     await rimraf(finalDir);
     await copyDir(output, path.resolve(__dirname, '..'));
     await rimraf(output);
+  } else if (mode === ALPHA_CORE_API) {
+    // Image builder types need to be fixed before they can be moved to their final location
```
celdrake: Comments in this block refer to `imageBuilder` and should refer to `alpha`

libs/ui-components/src/components/Catalog/InstallWizard/steps/SpecificationsStep.tsx
```diff
@@ -0,0 +1,216 @@
...
+              <FormGroup label={t('Target type')} isRequired />
+              <Stack hasGutter>
+                <StackItem>
+                  <RadioField
+                    id="fleet-radio"
+                    name="target"
+                    checkedValue="fleet"
+                    label={t('Existing Fleet')}
+                    description={t('Install to all devices in a fleet')}
+                    isDisabled={!canEditFleet || !canListFleet || !fleets.length}
```
celdrake: Minor, for later: Should have a "disabled" reason. (Also for the "Existing device" radio)

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
```diff
@@ -0,0 +1,139 @@
...
+      );
+    } else {
+      return (
+        <Alert isInline variant="warning" title={t('Existing OS image detected')}>
+          {t('')}
+        </Alert>
+      );
+    }
+  } else if (values.target === 'device') {
+    if (!values.device?.spec?.os?.image) {
```
celdrake, line 90: When choosing to deploy an application on a fleetless device without an OS in its spec, the banner is displayed and this is not correct.

I suspect the same would happen if I try to deploy an app  to a fleet that does not specify the OS yet.

libs/ui-components/src/components/Catalog/InstallWizard/steps/AppConfigStep.tsx
```diff
@@ -0,0 +1,202 @@
...
+    };
+
+    return { onAssetSelected, onBeforeArrayItemRemoved, onAssetCleared, selectedAssets: values.selectedAssets };
+  }, [values.selectedAssets, setFieldValue]);
+
+  return (
+    <Stack hasGutter>
+      <StackItem>
+        <FormGroupWithHelperText
+          label={t('Application name')}
```
celdrake, line 75: For later: validations for app name are missing. Should reuse the validations we have for other appTypes.

Eg. using "test*" results in a very large 400 error from Backend.

Also, when "deploy" fails and we go back, the previous error should be cleared (we also do this in theory in the other wizards).

libs/ui-components/src/components/common/CodeEditor/YamlEditorBase.tsx
```diff
@@ -21,15 +21,17 @@ loader.config({ monaco });
...
   onSave?: (yamlContent: string | undefined) => Promise<void>;
   isSaving: boolean;
   disabledEditReason?: string;
   editorRef?: React.MutableRefObject<monacoEditor.editor.IStandaloneCodeEditor | null>;
   readOnly?: boolean;
+  showActions?: boolean;
+  onChange?: (val: string) => void;
 };
 
 const YamlEditorBase = ({
```
celdrake, line 37: For some reason the YamlEditor has shrink in height to about 50% instead of ~100%.
If we can't get it fixed in time, we should reopen the previous bug, or create a new bug to address it later.

libs/ui-components/src/components/Catalog/CatalogItemCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+export type CatalogItemProps = {
+  catalogItem: CatalogItem;
+  onSelect: VoidFunction;
+};
+
+const CatalogItem: React.FC<CatalogItemProps> = ({ catalogItem, onSelect }) => {
+  const { t } = useTranslation();
+  return (
+    <Card isCompact isClickable>
+      <CardHeader
```
celdrake, line 29: (For later) There are some missing props as there are console errors related to each card.

```
CardHeader.js:30 Card: Clickable-only cards must have either the selectableActions.selectableActionAriaLabel or selectableActions.selectableActionAriaLabelledby prop passed in order to provide an accessible name to the clickable element.
```

libs/ui-components/src/components/Catalog/InstalledSoftware.tsx
```diff
@@ -0,0 +1,399 @@
...
+                      )}
+                    </React.Fragment>
+                  );
+                })}
+              </Tbody>
+            </Table>
+          )}
+        </CardBody>
+      </Card>
+      {deleteOs && (
```
celdrake, line 344: (Can be done later) IMO the delete modal, specially that of the Operating System, should indicate the consequences clearly.

libs/ui-components/src/components/Device/DeviceDetails/DeviceFleet.tsx
```diff
@@ -78,7 +77,7 @@ const DeviceFleet = ({ device }: { device?: Device }) => {
     return '-';
   }
 
-  const multipleOwnersCondition = getCondition(device.status?.conditions, ConditionType.DeviceMultipleOwners);
+  const multipleOwnersCondition = false;
```
celdrake: Needs revert?

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
```diff
@@ -0,0 +1,392 @@
...
+                  />
+                </FlexItem>
+                <FlexItem>
+                  <Stack>
+                    <StackItem>
+                      <Title headingLevel="h3">{assetItem.spec.displayName || assetItem.metadata.name}</Title>
+                    </StackItem>
+                    <StackItem>
+                      <Content component={ContentVariants.small}>
+                        Version: {selectedAssets[volumeIndex].assetVersion}, Channel{' '}
```
celdrake: Missing translations.

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
```diff
@@ -0,0 +1,392 @@
...
+                  onAssetCleared(volumeIndex);
+                  onChange({ ...value, reference: undefined });
+                }}
+              />
+            </SplitItem>
+          </Split>
+        ) : (
+          <Split hasGutter>
+            <SplitItem isFilled>
+              <TextInput
```
celdrake: Field is not marked as `required` from the schema, so it can be attempted to be edited and will fail on the API call.

libs/ui-components/src/components/Catalog/CatalogPage.tsx
```diff
@@ -0,0 +1,307 @@
...
+          )}
+        </Stack>
+      </EmptyStateBody>
+      {!hasFilters && (
+        <EmptyStateFooter>
+          <EmptyStateActions>
+            <Button
+              variant="link"
+              component="a"
+              href="https://docs.flightctl.io"
```
celdrake, line 76: Linked page does not exist. 

Should we keep `useAppLinks` and keep all links to the upstream github repo for now?
Also best to reuse `LearnMoreLink`

libs/ui-components/src/components/Catalog/CatalogPage.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+  const [catalogItems, isLoading, error, pagination, isUpdating] = useCatalogItems(catalogFilter);
+
+  const item = selectedItem
+    ? catalogItems?.find(
+        ({ metadata }) => metadata.name === selectedItem.itemName && metadata.catalog === selectedItem.catalog,
+      )
+    : undefined;
+
+  const filterIsEmpty = !catalogFilter.itemType;
```
celdrake: The empty state vs filter empty check is not working properly.

```suggestion
  const filterIsEmpty = catalogFilter.itemType.length === 0;
```

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
```diff
@@ -0,0 +1,139 @@
...
+const UpdateAlerts = () => {
+  const { t } = useTranslation();
+  const values = useFormikContext<InstallOsFormik | InstallAppFormik>().values;
+
+  if (values.target === 'fleet') {
+    if (!values.fleet?.spec.template.spec.os?.image) {
+      return (
+        <Alert isInline variant="warning" title={t('Fleet update')}>
+          {t(
+            'This will update the OS image for all (number) devices in the <name> fleet. Devices will download and apply the update according to the configured update policies.',
```
celdrake: Missing placeholders in this message. Either remove them or pass the relevant parameters.

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
```diff
@@ -0,0 +1,139 @@
...
+          )}
+        </Alert>
+      );
+    } else {
+      return (
+        <Alert isInline variant="warning" title={t('Existing OS image detected')}>
+          {t('')}
```
celdrake: Missing message. (also for the device branch later in the file)

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
```diff
@@ -0,0 +1,139 @@
...
+        </StackItem>
+        <StackItem>
+          <Card>
+            <CardTitle>{t('Target')}</CardTitle>
+            <CardBody>
+              <DescriptionList>
+                <DescriptionListGroup>
+                  <DescriptionListTerm>{t('Target type')}</DescriptionListTerm>
+                  <DescriptionListDescription>
+                    {values.target === 'fleet'
```
celdrake: Could we do this at the top before the render with a switch to also calculate `targetLabel`?

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
```diff
@@ -0,0 +1,342 @@
...
+              {devices.map((device, index) => (
+                <EnrolledDeviceTableRow
+                  key={device.metadata.name || ''}
+                  device={device}
+                  onRowSelect={(device) => () => handleDeviceSelect(device)}
+                  isRowSelected={isDeviceSelected}
+                  rowIndex={index}
+                  canEdit={false}
+                  canDecommission={false}
+                  decommissionAction={noopListAction}
```
celdrake: Why not define the `decommissionAction / resumeAction` as optional instead?

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
```diff
@@ -0,0 +1,342 @@
...
+  );
+};
+
+const getArtifactLabel = (artifact: CatalogItemArtifact, t: TFunction) => {
+  if (artifact.name) {
+    return artifact.name;
+  }
+  switch (artifact.type) {
+    case CatalogItemArtifactType.CatalogItemArtifactTypeQcow2:
+      return t('OpenShift Virtualization ({{type}})', { type: CatalogItemArtifactType.CatalogItemArtifactTypeQcow2 });
```
celdrake: Code would be shorter if you extracted `type` as a variable and used that in all cases expect for `default`

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
```diff
@@ -0,0 +1,342 @@
...
+      return t('Anaconda Installer ({{type}})', { type: CatalogItemArtifactType.CatalogItemArtifactTypeAnacondaIso });
+    case CatalogItemArtifactType.CatalogItemArtifactTypeGce:
+      return t('Google Cloud ({{type}})', { type: CatalogItemArtifactType.CatalogItemArtifactTypeGce });
+    case CatalogItemArtifactType.CatalogItemArtifactTypeRaw:
+      return t('KVM/custom cloud import ({{type}})', { type: CatalogItemArtifactType.CatalogItemArtifactTypeRaw });
+    case CatalogItemArtifactType.CatalogItemArtifactTypeVhd:
+      return t('Microsoft Hyper-V ({{type}})', { type: CatalogItemArtifactType.CatalogItemArtifactTypeVhd });
+    case CatalogItemArtifactType.CatalogItemArtifactTypeVmdk:
+      return t('VMware vSphere ({{type}})', { type: CatalogItemArtifactType.CatalogItemArtifactTypeVmdk });
+    default:
```
celdrake, line 230: Since both `type` and `name` are optional, I think it would be better to have the `default` case be `Unknown type` rather than fallback to the last known type.

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
```diff
@@ -0,0 +1,342 @@
...
+          <StackItem>
+            <Button
+              component="a"
+              variant="link"
+              isInline
+              iconPosition="end"
+              icon={<ExternalLinkAltIcon />}
+              target="_blank"
+              rel="noopener noreferrer"
+              href={artifactUrl}
```
celdrake: Urls are being treated as relative (At least with the sample catalog)
You could use `ImageUrl` and it would handle converting the link to absolute

libs/ui-components/src/components/Catalog/InstallWizard/steps/SpecificationsStep.tsx
```diff
@@ -0,0 +1,216 @@
...
+        </GridItem>
+        {!hideReadmeLink && !!currentVersion?.readme && (
+          <GridItem span={4}>
+            <Button onClick={() => setShowReadme(true)} variant="link">
+              {t('Show readme')}
+            </Button>
+          </GridItem>
+        )}
+      </Grid>
+      {showReadme && currentVersion?.readme && (
```
celdrake, line 142: (Can be done later) Could we also show the `deprecation` messages?

libs/ui-components/src/components/Device/DeviceDetails/DeviceFleet.tsx
rawagner: yes, reverted

libs/types/models/ConditionType.ts
rawagner: removed

libs/ui-components/src/components/Catalog/InstallWizard/steps/SpecificationsStep.tsx
rawagner: just added the deprecation msgs/labels to catalog page, installed software, version dropdowns, catalog item details and also upgrade modal

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
rawagner: Perfect, Im using `ImageUrl` now

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
rawagner: Ok, updated

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
rawagner: Updated, the type is added afterwards

libs/ui-components/src/components/Catalog/InstallWizard/steps/SelectTargetStep.tsx
rawagner: you are right, changed to optional (also related props like `canDecomission` etc)

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
rawagner: yes, in our example schema, the field is not marked as required - which is actually incorrect. There is nothing we can do about that in the UI however IMO. It is a responsibility of a catalog item creator to have a correct schema, otherwise the form/editor will allow/disallow submitting invalid config.

celdrake: I was referring to the  "NGinx Edge proxy" app which seems to have the field defined as required.
```
          volumes:
            type: array
            items:
              type: object
              required: ["name", "image"]  // is this the correct way to define the "required" fields?
              properties: 
                name:
                  type: string
                  description: "Volume name"
                image:
                  type: object
                  description: "Volume iamge"
                  properties:
                    reference:
                      type: string

```

libs/ui-components/src/components/Catalog/UpdateModal/OsUpdateModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const handleSelectionChange = React.useCallback((_nodeId: string, version: string) => {
+    setSelectedVersion(version);
+  }, []);
+
+  const handleUpdate = async () => {
+    setIsUpdating(true);
+    try {
+      await onUpdate(selectedVersion, currentChannel);
+      onClose();
+    } catch (e) {
```
celdrake: In `ResourceCatalogPage` we currently only do an action if the `catalogItemVErsion` is found. And here we ignore any possible error we may receive from the API calls, etc.

IMO we should handle both and show any error that may ocurr.

libs/ui-components/src/components/Catalog/CatalogItemDetails.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+type CatalogItemDetailsProps = CatalogItemDetailsPanelProps & {
+  onInstall: (installItem: { item: CatalogItem; channel: string; version: string }) => void;
+};
+
+const getPageContentTop = () => {
+  // Try multiple selectors to find the masthead
+  const masthead =
+    document.getElementById('stack-inline-masthead') ||
+    document.querySelector('.pf-v5-c-masthead') ||
```
celdrake: Main tag is `<header class="pf-v6-c-masthead pf-m-display-inline" id="stack-inline-masthead">`

The v5 class won't be found for sure. So I'd keep just the first selector.

libs/ui-components/src/components/Catalog/CatalogItemDetails.tsx
```diff
@@ -0,0 +1,279 @@
...
+          </Content>
+        ) : (
+          t('N/A')
+        )}
+      </GridItem>
+    </Grid>
+  );
+};
+
+export const getDefaultChannelAndVersion = (item: CatalogItem) => {
```
celdrake, line 257: Not for now, but it may  be better to do the opposite (meaning figure out if possible which is the latest version, and select the channel holding that version).

libs/ui-components/src/components/Catalog/utils.ts
```diff
@@ -0,0 +1,310 @@
...
+  } else if (newOsImage && currentOsImage !== newOsImage) {
+    appendJSONPatch({
+      path: `${specPath}spec/os/image`,
+      patches: allPatches,
+      newValue: newOsImage,
+      originalValue: currentOsImage,
+    });
+  }
+
+  const newLabels = fromAPILabel({
```
celdrake, line 175: Can we be removing the OS image in this function?
If not, I'd remove the branch for removal. Otherwise it's confusing as it seems we are adding the OS labels even in that case.

libs/ui-components/src/components/Device/DevicesPage/useDevices.ts
```diff
@@ -71,7 +80,9 @@ const getDevicesEndpoint = ({
   if (summaryOnly) {
     params.set('summaryOnly', 'true');
   }
-  if (nextContinue !== undefined) {
+  if (limit !== undefined) {
+    params.set('limit', `${limit}`);
+  } else if (nextContinue !== undefined) {
```
celdrake: Can we combine this with the statement in the next "if"?
I'm finding this a bit confusing TBH.

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
```diff
@@ -0,0 +1,392 @@
...
+type VolumeImageValue = {
+  reference?: string;
+};
+
+/**
+ * Extract the volume index from the field ID.
+ * Field ID format: "root_volumes_0_image" -> extracts index 0
+ */
+const getVolumeIndexFromId = (fieldId: string): number => {
+  const match = fieldId.match(/_(\d+)_image$/);
```
celdrake: I saw in another place you used the full `root_volumes...` regexp.
Is the pattern the same here? If so, I'd use it as well so it can be easily tracked together with the other places it's used.

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
rawagner: the alert msgs are now updated

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
rawagner: the alert msgs are now updated

apps/standalone/src/app/routes.tsx
rawagner: updated

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
rawagner: these were mostly placeholders..the alerts are now updated and should be correct

libs/ui-components/src/components/Catalog/CatalogItemCard.tsx
rawagner: fixed

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
rawagner: fixed

libs/ui-components/src/components/Catalog/InstallWizard/steps/AppConfigStep.tsx
rawagner: fixed

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
rawagner: ok, I misunderstood. The volume image is now properly marked as required

libs/ui-components/src/components/Catalog/UpdateModal/OsUpdateModal.tsx
rawagner: fixed

libs/ui-components/src/components/Catalog/utils.ts
rawagner: no, we cannot remove. Fixed

libs/ui-components/src/components/Device/DevicesPage/useDevices.ts
rawagner: hmm, no I dont think we can combine with the next if.
We could do
```
if (limit !== undefined || nextContinue !== undefined) {
    params.set('limit', `${limit ?? PAGE_SIZE}`);
}
if (nextContinue) {
  params.set('continue', nextContinue);
}
```

thought that does not improve the readability IMO. Any suggestions ?

rawagner: actually this was some leftover. I am not using limit anywhere. Removed.

libs/ui-components/src/components/DynamicForm/VolumeImageField.tsx
rawagner: updated, the regex is also reused across components

libs/ui-components/src/components/Catalog/CatalogItemDetails.tsx
rawagner: ok, now we do semver sort to get latest version
then pick the latest release, or prerelease (if there is no release version)

libs/ui-components/src/components/Catalog/CatalogItemDetails.tsx
rawagner: selector is now improved
works with both standalone and ocp console
as a backup I've added selector for the toolbar that contains the `copy login cmd` etc

apps/ocp-plugin/console-extensions.json
```diff
@@ -35,6 +35,16 @@
       "section": "fctl"
     }
   },
+  {
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "fctl-catalog",
+      "name": "%plugin__flightctl-plugin~Software Catalog%",
```
celdrake, line 42: Are we going for the "Software Catalog" naming in the end?

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const UpdateAlerts = ({ catalogItem }: UpdateAlertsProps) => {
+  const { t } = useTranslation();
+  const { values } = useFormikContext<InstallOsFormik>();
+
+  const osImageName = `${catalogItem.spec.displayName || catalogItem.metadata.name}:${values.version}`;
+
+  if (values.target === 'fleet') {
+    const numOfDevices = `${values.fleet?.status?.devicesSummary?.total || 0}`;
+    if (!values.fleet?.spec.template.spec.os?.image) {
```
celdrake, line 56: I think we should check the `catalogItem.type` and only show this warning when it's an OS that's being added.

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
celdrake: The issue that we're deploying an application to a device that didn't specify the OS is still present.
Same as before, I think we need to check that the `catalogItem.type` is for an OS.

libs/ui-components/src/components/Catalog/InstallWizard/steps/SpecificationsStep.tsx
rawagner: added disabled reason

libs/ui-components/src/components/Catalog/InstallWizard/steps/ReviewStep.tsx
rawagner: we have the `isOs` prop and only then render the alerts..but you are right that we can do that based on the `catalogItem` prop.
Fixed

apps/ocp-plugin/console-extensions.json
rawagner: I think so. We discussed in the past that `Software Catalog` is widely used term.

apps/standalone/src/app/routes.tsx
```diff
@@ -215,6 +231,15 @@ const getAppRoutes = (t: TFunction): ExtendedRouteObject[] => [
           </TitledRoute>
         ),
       },
+      {
+        path: 'catalog/:fleetId/:catalogId/:itemId',
```
celdrake, line 235: The pattern here is a bit weird as for all the other fleet routes we have
`fleet/<fleetId>/subpath`

I assume this is necessary because we want to have this page not go via the usual `FleetDetails` route, so it's fine.

libs/ui-components/src/components/Catalog/EditWizard/steps/UpdateStep.tsx
```diff
@@ -0,0 +1,138 @@
...
+                  <FormGroup label={t('Target version')} />
+                </GridItem>
+                <GridItem span={5}>{currentVersion.version}</GridItem>
+                <GridItem span={2}>
+                  <Icon>
+                    <ArrowRightIcon />
+                  </Icon>
+                </GridItem>
+                <GridItem span={5}>
+                  {updates.length ? (
```
celdrake, line 90: Can be done later, but I think it should be possible for users to rollback to a previous version.
Now, if they're at the latest version, then they just see "Up to date" without an option to change it.

libs/ui-components/src/components/Catalog/useCatalogs.ts
```diff
@@ -0,0 +1,106 @@
...
+    const categories: CatalogItemCategory[] = [];
+    let types = itemType ? [...itemType] : [];
+
+    if (appTypeIds.every((id) => types.includes(id))) {
+      categories.push(CatalogItemCategory.CatalogItemCategoryApplication);
+      types = types.filter((t) => !appTypeIds.includes(t));
+    }
+
+    if (categories.length === 1) {
+      parts.push(`spec.category==${categories[0]}`);
```
celdrake, line 46: Nit: I think it should work with the plural form even if only one category is selected, and same for the other filters.

```
flightctl get catalogitems --field-selector="spec.category in (aplication)" 
```

celdrake (approved):

PR #544, opened 2026-02-25
EDM-3419 Allow OCI registries with subpath

celdrake:
Pending final adjustments on the Backend to allow the same behaviour.


celdrake:
Closed as not a bug, the subpath is part of the image, not the base repository URI.

PR #547, opened 2026-02-25
EDM-3383: Refresh device before displaying YAML

celdrake:
When the YAML editor opens, we now refresh the device details so that the editor shows the most updated version.

This helps for example when quickly switching from the Terminal tab to the YAML tab, as the device annotations are updated.

## Summary by CodeRabbit

* **New Features**
  * Added a brief loading state and delayed refresh when opening the YAML editor on device details pages for smoother loading.
  * YAML editor now conditionally performs an initial refresh when a device has active console sessions.

* **Style / Layout**
  * Reduced the YAML editor minimum height and adjusted default editor and terminal tab heights for improved fit.

* **Localization**
  * Added an English translation for "Refreshing details".

### DIFF
--- a/libs/i18n/locales/en/translation.json
+++ b/libs/i18n/locales/en/translation.json
@@ -504,6 +504,7 @@
   "The current YAML is invalid. Fix the errors, or click ʼReloadʼ to discard your changes.": "The current YAML is invalid. Fix the errors, or click ʼReloadʼ to discard your changes.",
   "This object has been updated.": "This object has been updated.",
   "Click reload to see the new version.": "Click reload to see the new version.",
+  "Refreshing details": "Refreshing details",
   "Copy to clipboard": "Copy to clipboard",
   "Content copied to clipboard": "Content copied to clipboard",
   "Download yaml": "Download yaml",
--- a/libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
+++ b/libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Alert, AlertActionCloseButton, Stack, StackItem } from '@patternfly/react-core';
+import { Alert, AlertActionCloseButton, Bullseye, Spinner, Stack, StackItem } from '@patternfly/react-core';
 import { CodeEditorProps as PfCodeEditorProps } from '@patternfly/react-code-editor';
 import { dump, load } from 'js-yaml';
 import { compare } from 'fast-json-patch';
@@ -7,6 +7,7 @@ import type * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api';
 
 import { AuthProvider, Device, Fleet, PatchRequest, Repository, ResourceKind } from '@flightctl/types';
 import { ImageBuild } from '@flightctl/types/imagebuilder';
+import { showSpinnerBriefly } from '../../../utils/time';
 import { fromAPILabel } from '../../../utils/labels';
 import { getLabelPatches } from '../../../utils/patch';
 import { getErrorMessage, isResourceVersionTestFailure } from '../../../utils/error';
@@ -107,6 +108,8 @@ const createPatchList = (original: FlightCtlYamlResource, updated: FlightCtlYaml
   // To prevent that, we'll use a fallback that patches the top-level mutable fields that were modified.
   return getFallbackPatches(original, updated);
 };
+
+const YAML_TAB_REFETCH_DELAY_MS = 500;
 
 const YamlEditor = <R extends FlightCtlYamlResource>({
   apiObj,
@@ -257,6 +260,39 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
       )}
     </div>
   );
+};
+
+/**
+ * Use YamlEditorLoader instead of the default YamlEditor component to trigger a refetch of the resource before rendering the editor.
+ *
+ * @param props - YamlEditor component props.
+ * @returns A spinner while the resource is being refetched, then the YamlEditor component.
+ */
+export const YamlEditorLoader = <R extends FlightCtlYamlResource>(props: YamlEditorProps<R>) => {
+  const { t } = useTranslation();
+  const [initialLoad, setInitialLoad] = React.useState(false);
+  const { refetch } = props;
+
+  React.useEffect(() => {
+    const forceRefresh = async () => {
+      setInitialLoad(false);
+      // Wait before triggering the refetch to give time for the resource to be updated.
+      await showSpinnerBriefly(YAML_TAB_REFETCH_DELAY_MS);
+      refetch();
+      setInitialLoad(true);
+    };
+    void forceRefresh();
+  }, [refetch]);
+
+  if (!initialLoad) {
+    return (
+      <Bullseye>
+        <Spinner size="lg" aria-label={t('Refreshing details')} />
+      </Bullseye>
+    );
+  }
+
+  return <YamlEditor {...props} />;
 };
 
 export default YamlEditor;
--- a/libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
+++ b/libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
@@ -109,7 +109,8 @@ const createPatchList = (original: FlightCtlYamlResource, updated: FlightCtlYaml
   return getFallbackPatches(original, updated);
 };
 
-const YAML_TAB_REFETCH_DELAY_MS = 500;
+const YAML_TAB_REFETCH_DELAY_BEFORE_MS = 400;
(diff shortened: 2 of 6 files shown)


libs/i18n/locales/en/translation.json
```diff
@@ -323,6 +323,7 @@
   "System image mismatch": "System image mismatch",
   "Desired system image: {{ desiredOsImage }}": "Desired system image: {{ desiredOsImage }}",
   "System services": "System services",
+  "Refreshing device": "Refreshing device",
```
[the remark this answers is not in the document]
celdrake: The file was generated using `npm run i18n`, not edited manually.

libs/ui-components/src/components/Device/DeviceDetails/DeviceYamlTab.tsx
```diff
@@ -0,0 +1,48 @@
...
+    // Since refetch is async, we add an extra wait so the YAML editor opens with the latest device version
+    // Otherwise, the user would need to click
+    const fetchDevice = async () => {
+      refetch();
+      setReady(false);
+      await showSpinnerBriefly(YAML_TAB_REFETCH_DELAY_MS);
+      setReady(true);
+    };
+    void fetchDevice();
+  }, [refetch]);
```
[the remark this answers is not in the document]
celdrake: Fixed comment.
The implementation with "isUpdating" is not working as expected. The request is very fast, so 500ms should be more than enough.

rawagner (commented):
Shouldn't we do this for any Yaml view ?
the `YamlEditor` already accepts `refresh`, so you could move the logic directly to `YamlEditor`.

celdrake:
> Shouldn't we do this for any Yaml view ? the `YamlEditor` already accepts `refresh`, so you could move the logic directly to `YamlEditor`.

Seems like a good idea. However, I'm having some issues with the fix working consistently (I suspect the same thing happens with my original PR anyway). 

The problem is that the Backend does not remove the annotations on the device that indicate there is an open console immediately after the connection is closed. It's not infrequent for the UI to load a device version that will become obsolete in a very short time (even when adding a delay of ~700 ms).

I'm trying to provide a better fix.

celdrake:
@rawagner I moved the code per your suggestion so it's in the YAML editor. I realized that the wait should be done BEFORE triggering refetch, so the device has time to be updated after the console session is closed (meaning the annotations are removed)

Do you think we should add this delay only for devices, or for all resources? In my experience, it's not needed for other resources, as this is very specific to devices and how the console sessions modify the device `resourceVersion`, so I'd prefer not to add the extra delay everywhere.

In any case, If we want it for all resources, the change is to only export `YamlEditorLoader` as the default component.

rawagner:
> @rawagner I moved the code per your suggestion so it's in the YAML editor. I realized that the wait should be done BEFORE triggering refetch, so the device has time to be updated after the console session is closed (meaning the annotations are removed)
> 
> Do you think we should add this delay only for devices, or for all resources? In my experience, it's not needed for other resources, as this is very specific to devices and how the console sessions modify the device `resourceVersion`, so I'd prefer not to add the extra delay everywhere.
> 
> In any case, If we want it for all resources, the change is to only export `YamlEditorLoader` as the default component.

+1 for Adding the wait only for devices

Also I wonder - should we check if the device has the console session annotation, if yes, only then trigger the wait ?

libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
```diff
@@ -259,4 +262,37 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
...
+  React.useEffect(() => {
+    const forceRefresh = async () => {
+      setInitialLoad(false);
+      // Wait before triggering the refetch to give time for the resource to be updated.
+      await showSpinnerBriefly(YAML_TAB_REFETCH_DELAY_MS);
+      refetch();
+      setInitialLoad(true);
+    };
+    void forceRefresh();
+  }, [refetch]);
```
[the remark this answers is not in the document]
celdrake: The problem happens when the user moves from the Terminal tab (closes a device console session) quickly to the YAML tab, it takes some time for the backend to update the device to have its "device-controller/console" annotation removed.

If the update happens after we mount the YamlEditor, then the resourceVersion with the annotation becomes "locked in" the YAML editor and the user then needs to click "Reload" before they are able to modify the YAML.

Performing the refetch after the delay allows the device to become updated with the removed annotation, therefore allows us to load the YAML editor with the latest available version from the beginning.

celdrake:
@rawagner Yes, also a good idea. Please check the second commit with the changes.

libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
```diff
@@ -259,4 +262,57 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
...
+/**
+ * Use YamlEditorLoader instead of the default YamlEditor component to trigger a conditional fetch of the resource before rendering the editor.
+ * The refetch is performed only if `shouldFetchInitially` returns true.
+ *
+ * @param props - YamlEditor component props plus `shouldFetchInitially`.
+ * @returns A spinner while the resource is being refetched, then the YamlEditor component.
+ */
+export const YamlEditorLoader = <R extends FlightCtlYamlResource>(
+  props: YamlEditorProps<R> & {
+    shouldFetchInitially: () => boolean;
```
rawagner: why is this a function instead of a simple boolean?

libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
```diff
@@ -259,4 +262,57 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
...
+  const shouldFetchInitiallyRef = React.useRef(shouldFetchInitially);
+  shouldFetchInitiallyRef.current = shouldFetchInitially;
+
+  const [isLoaded, setIsLoaded] = React.useState(() => !props.shouldFetchInitially());
+
+  React.useEffect(() => {
+    let cancelled = false;
+
+    const shouldFetch = shouldFetchInitiallyRef.current();
+    if (shouldFetch) {
```
rawagner, line 290: if you keep the function, should you also execute it ? 
```suggestion
    if (shouldFetch()) {
```


edit: my bad, i didnt notice you actually do a line above :)

libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
celdrake: That was to avoid evaluating the device annotation on each render. With a function, we'd only evaluate it whenever the YAML is open. Or perhaps we could make it a memo.

What do you think is best?

rawagner: if there is a need, it should be a responsibility of a caller. So the YamlEditorLoader should simply accept a boolean prop.

Also IMO no need to memoize the access to annotations.

celdrake: Done, though there was a bug that `setIsLoaded` was being updated immediately without the wait.
I also split the delay between before and after the wait, as otherwise after triggering the refetch, the previous data would still be loaded.

And I made the YAML editor smaller so that whenever a new version is available, there's space for the "Refresh" banner to be clearly visible.

libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
```diff
@@ -259,4 +263,61 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
...
+      };
+      void doRefetch();
+    } else {
+      setIsLoaded(true);
+    }
+
+    return () => {
+      cancelled = true;
+    };
+  }, [refetch]);
```
rawagner: since we want to do fetch only on initial load, do we even need `[refetch]` dependency/React.ref ?


```suggestion

  const [isLoaded, setIsLoaded] = React.useState(!shouldFetchInitially);

  React.useEffect(() => {
    let cancelled = false;

    if (shouldFetchInitially) {
      const doRefetch = async () => {
        setIsLoaded(false);
        await showSpinnerBriefly(YAML_TAB_REFETCH_DELAY_BEFORE_MS);
        if (cancelled) {
          return;
        }
        refetch();
        // Give some time for refetch to complete before marking the editor as loaded
        await showSpinnerBriefly(YAML_TAB_REFETCH_DELAY_AFTER_MS);
        setIsLoaded(true);
      };
      void doRefetch();
    } else {
      setIsLoaded(true);
    }

    return () => {
      cancelled = true;
    };
  // run only on mount
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

celdrake: True, modified.

rawagner (approved):

PR #594, opened 2026-03-25
EDM-3580: Fix error 404 in device catalog for managed devices

celdrake:
To retrieve for the devices's owner fleet, we were attempting to match to a ResourceSync.

After the fix
device-catalog

## Summary by CodeRabbit

* **Refactor**
  * Consolidated owner-parsing into a shared utility and updated components to use it. Internal code reorganization only; no changes to visible UI or behavior.

### DIFF
--- a/libs/ui-components/src/utils/resource.ts
+++ b/libs/ui-components/src/utils/resource.ts
@@ -1,4 +1,16 @@
-import { ObjectMeta } from '@flightctl/types';
+import { ObjectMeta, ResourceKind } from '@flightctl/types';
+
+export const getOwnerName = (ownerKind: ResourceKind, owner: string | undefined): string | undefined => {
+  if (!owner) {
+    return undefined;
+  }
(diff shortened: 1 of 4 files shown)


libs/ui-components/src/utils/resource.ts
```diff
@@ -1,4 +1,16 @@
-import { ObjectMeta } from '@flightctl/types';
+import { ObjectMeta, ResourceKind } from '@flightctl/types';
+
+export const getOwnerName = (ownerKind: ResourceKind, owner: string | undefined): string | undefined => {
+  if (!owner) {
+    return undefined;
+  }
+  const prefix = `${ownerKind}/`;
```
rawagner, line 7: I dont think we even need `ownerKind`. We can just split by `/` as it has to be there exactly once. It cannot be part of Kind name, nor owner resource name

rawagner (approved):

PR #601, opened 2026-03-29
EDM-2796: Redirect url is not calculated correctly

ori-amizur:
## Summary by CodeRabbit

* **New Features**
  * Login/logout flows now accept and persist a per-request redirect origin so users return to the intended UI origin.

* **Security**
  * Redirect target origin is strictly validated against the incoming request to prevent improper redirects.
  * New optional config to trust forwarded headers (with trusted proxy CIDR support) for correct origin resolution behind proxies.

* **Documentation**
  * Backend config docs updated to document new flags for forwarded-header trust and trusted proxy CIDRs.

* **Tests**
  * Unit tests added for redirect origin resolution and trusted-proxy logic.

### DIFF
--- a/proxy/auth/redirect.go
+++ b/proxy/auth/redirect.go
@@ -1,0 +1,204 @@
+package auth
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/flightctl/flightctl-ui/config"
+)
+
+const oauthRedirectURICookiePrefix = "oauth_redirect_uri_"
+
+// oauthCallbackFromOrigin builds the OAuth redirect_uri (…/callback) for the given UI origin
+// (scheme + host, no path). The path prefix comes from config.BaseUiUrl so deployments served
+// from a subpath (e.g. https://host/ui) resolve to …/ui/callback instead of …/callback.
+func oauthCallbackFromOrigin(origin *url.URL) (string, error) {
+	baseUI, err := url.Parse(config.BaseUiUrl)
+	if err != nil {
+		return "", fmt.Errorf("invalid BASE_UI_URL configuration: %w", err)
+	}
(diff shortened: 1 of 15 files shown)


proxy/config/config.go
```diff
@@ -19,8 +21,99 @@ var (
...
 )
 
+// trustedProxyNets is parsed from TRUSTED_PROXY_CIDRS (comma-separated). When non-empty and
+// TrustXForwardedHeaders is true, forwarded headers apply only when the immediate client IP
+// (r.RemoteAddr) falls within one of these networks. When empty, all clients are accepted
+// once TrustXForwardedHeaders is true (use only if the proxy is not reachable from untrusted clients).
+var trustedProxyNets []*net.IPNet
+
+func init() {
+	trustedProxyNets = parseTrustedProxyCIDRs(getEnvVar("TRUSTED_PROXY_CIDRS", ""))
```
rawagner: we should also make possible to set these new env variables via helm chart/quadlet config.

We should also document these new variables in CONFIGURATION.md

ori-amizur: Setting via helm chart/quadlet config will be set in another PR on flightctl repo.
Added documentation

proxy/auth/redirect.go
```diff
@@ -0,0 +1,150 @@
...
+		Name:     cookieName,
+		Value:    redirectURI,
+		Secure:   config.TlsCertPath != "",
+		HttpOnly: true,
+		SameSite: http.SameSiteLaxMode,
+		Path:     "/",
+		MaxAge:   600,
+	}
+	http.SetCookie(w, &cookie)
+}
```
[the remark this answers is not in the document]
ori-amizur: Your suggested improvement (config.TlsCertPath != "") doesn't seem to have any affect

rawagner (approved):

PR #602, opened 2026-03-30
EDM-3608: Check permissions for catalog/item list

rawagner:
## Summary by CodeRabbit

* **New Features**
  * Permission-based access gating added across catalog pages, wizards, install/edit flows so content is shown only when appropriate list/get/create/patch rights are present.
  * "Choose from catalog" button now appears only for users with catalog-item list permission.
  * Pages and wizards display loading indicators while permissions are being verified to improve UX during access checks.


libs/ui-components/src/components/Catalog/ResourceCatalog/ResourceCatalogPage.tsx
```diff
@@ -20,6 +23,11 @@ type ResourceCatalogPageProps = {
   onInstall: (installItem: { item: CatalogItem; channel: string; version: string }) => void;
 };
 
+const catalogPagePermissions = [
```
celdrake, line 26: This is good for CatalogPage, but I believe other sections in the UI also need additional checks.
Quick review from Cursor:

```
The add/edit item wizard’s General Info step, install wizard, edit-from-catalog wizard, edit item wizard, and asset picker modal can still issue catalog/catalog-item requests without a matching client-side permission check first;
```

rawagner: added those additional locations as well

siserafin (approved):
lgtm

PR #625, opened 2026-04-20
EDM-3687: Allow user to go back to step to fix problems

celdrake:
Fixes the issue in the CreateFleetWizard whereby when a validation error was present in step N, and user moved to N-1, then they couldn't move back to step N.

### DIFF
--- a/libs/ui-components/src/components/Fleet/CreateFleet/CreateFleetWizard.tsx
+++ b/libs/ui-components/src/components/Fleet/CreateFleet/CreateFleetWizard.tsx
@@ -57,9 +57,12 @@ const getValidStepIds = (formikErrors: FormikErrors<FleetFormValues>): string[] 
   return validStepIds;
 };
 
+// Do not disable the current step where there could be validation errors
 const isDisabledStep = (stepId: string, validStepIds: string[]) => {
-  const validIndex = validStepIds.indexOf(stepId);
-  return validIndex === -1 || validIndex !== orderedIds.indexOf(stepId);
+  const stepIdx = orderedIds.findIndex((orderedStepId) => orderedStepId === stepId);
+  return orderedIds.some((orderedId, orderedStepIdx) => {
+    return orderedStepIdx < stepIdx && !validStepIds.includes(orderedId);
+  });
 };
 
 const CreateFleetWizard = () => {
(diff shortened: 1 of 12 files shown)


libs/ui-components/src/components/Fleet/CreateFleet/CreateFleetWizard.tsx
```diff
@@ -57,9 +57,12 @@ const getValidStepIds = (formikErrors: FormikErrors<FleetFormValues>): string[]
...
 };
 
+// Do not disable the current step where there could be validation errors
 const isDisabledStep = (stepId: string, validStepIds: string[]) => {
-  const validIndex = validStepIds.indexOf(stepId);
-  return validIndex === -1 || validIndex !== orderedIds.indexOf(stepId);
+  const stepIdx = orderedIds.findIndex((orderedStepId) => orderedStepId === stepId);
+  return orderedIds.some((orderedId, orderedStepIdx) => {
+    return orderedStepIdx < stepIdx && !validStepIds.includes(orderedId);
+  });
```
rawagner: the condition is quite confusing TBH.
Can't we just pass a current step ID to this function and return early ?


```suggestion
  const isDisabledStep = (currentStepId: string, stepId: string, validStepIds: string[]) => {
    if (currentStepId === stepId) {
      return false; //current step is never disabled
    }
    const validIndex = validStepIds.indexOf(stepId);
    return validIndex === -1 || validIndex !== orderedIds.indexOf(stepId);
  }
```

also we use this condition in multiple wizards - maybe its worth making it a common func.

celdrake: With this change, the initial bug would not be fixed:  When on step 2 with some invalid input, users can go back to step 1. And now clicking on "Next" does nothing since the step with the error is a "future step" from the POV of the new current step.

I agree that it would be better to unify the check for the disabled steps in the different Wizards.
I can see two options:

- Use in the CreateFleetWizard the pattern we have in the other wizards, that for step N+2, they check the steps N, N+1.
- Make this function reusable and use it in all the wizards.

Personally I find the second one better in the long run, even though it's harder to understand. We could properly document it and it would not require new custom logic for any new Wizards.

rawagner: Ok, sounds good. Lets go with the second one then.

celdrake: Thanks.
I think it should be clearer now.

rawagner (approved):

PR #630, opened 2026-04-24
EDM-3777: Filter repositories via the API

celdrake:
Filtering by name in Repositories page worked differently to the rest of the pages, since it was done client-side and used fuzzy search.

Now the filtering in Repositories / ResourceSyncs will be done via the API with "contains"


libs/ui-components/src/components/Repository/useRepositories.ts
```diff
@@ -38,7 +44,16 @@ export const useRepositories = (): [
...
   const { currentPage, setCurrentPage, itemCount, nextContinue, onPageFetched } = useTablePagination<RepositoryList>();
-  const [repoEndpoint, isDebouncing] = useRepositoriesEndpoint({ nextContinue });
+
+  const prevNameRef = React.useRef(name);
+  React.useEffect(() => {
+    if (prevNameRef.current !== name) {
+      prevNameRef.current = name;
+      setCurrentPage(1);
+    }
+  }, [name, setCurrentPage]);
```
rawagner, line 54: why are you using name Ref ? Would this be sufficient ?
```suggestion
  React.useEffect(() => {
      setCurrentPage(1);
  }, [name, setCurrentPage]);
```

celdrake: The only difference is that with the `ref` we avoid a re-render in the first mount.
I can remove it, since it's not very clear in any case.

rawagner (approved):

PR #640, opened 2026-05-15
EDM-3941: New version action on image build

rawagner:
requires https://github.com/flightctl/flightctl/pull/2961


rawagner:
included in https://github.com/flightctl/flightctl-ui/pull/643

PR #645, opened 2026-05-20
Add spacer to YamlEditor alert messages

celdrake:
There was no spacing between the row with the buttons and the alert messages in the Yaml Editor.

yaml-spacing

### DIFF
--- a/libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
+++ b/libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
@@ -201,65 +201,73 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
   };
 
   return (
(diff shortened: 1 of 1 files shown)


libs/ui-components/src/components/common/CodeEditor/YamlEditor.tsx
```diff
@@ -222,6 +222,7 @@ const YamlEditor = <R extends FlightCtlYamlResource>({
 
       {isSavedSuccessfully && (
         <Alert
+          className="fctl-yaml-editor__alert"
```
rawagner: instead of custom css, could we use Stack/StackItem ?

celdrake: Right, fixed now.

rawagner (approved):

PR #655, opened 2026-05-28
EDM-4012: Reset filters when logType selection changes

celdrake:
When log type changes, we must reset the selected filters since not all filters apply to all log types (eg. for "file path" only filter is the file name)


celdrake:
Moving to Draft as I'm making bigger changes and it may be better to add this fix on top of those changes

celdrake:
It's now incorporated in https://github.com/flightctl/flightctl-ui/pull/659

PR #661, opened 2026-06-01
EDM-4049: Do now show sync status of RS without repositories

celdrake:
If users delete the repository that a RS is associated with, the UI was still showing the Pending/Error notification about that RS status. 

Since eventually the RS will fail with a condition, and the resource sync status won't change while the repository is missing, we'll hide those notifications.

```
  status:
    conditions:
    - lastTransitionTime: "2026-06-01T07:49:41.454211613Z"
      message: Repository of name "test-repo" not found
      reason: repository resource not found
      status: "False"
      type: Accessible
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the shared `libs/ui-components/` ResourceSync status component.
  * Hides pending and error notifications when the referenced repository does not exist.
  * Suppresses notifications while repository data is loading or unavailable.
* The change affects shared UI behavior and can apply to both standalone and OCP plugin consumers.
* No changes affect `libs/types/`, `libs/i18n/`, `libs/cypress/`, platform-specific app code, the Go auth proxy, container builds, E2E tests, or CI configuration.

### DIFF
--- a/libs/ui-components/src/components/ResourceSync/ResourceSyncImportStatus.tsx
+++ b/libs/ui-components/src/components/ResourceSync/ResourceSyncImportStatus.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Alert, AlertActionCloseButton, PageSection, Stack, StackItem } from '@patternfly/react-core';
 import { Trans } from 'react-i18next';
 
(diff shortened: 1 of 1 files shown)


asmasarw (approved):
LGTM

PR #684, opened 2026-06-09
EDM-3805: Remove icon from new Catalog form

celdrake:
Remove the icon from the Catalog's form since it's not displayed anywhere.

## Summary by CodeRabbit

* **Shared UI / catalog creation flow (libs/ui-components/)**
  * Removed the catalog **Icon** field from the Catalog creation wizard/modal end-to-end:
    * Dropped the `IconUploadField` control from `CreateCatalogModal`.
    * Removed `icon` from the create form values type and from Formik/Yup `validationSchema`.
    * Stopped generating/including `spec.icon` in the catalog JSON patches and the `Catalog` payload.

* **i18n (libs/i18n/)**
  * Reordered/regrouped catalog-related translation entries in `libs/i18n/locales/en/translation.json` (including the `Icon` label within that block) and moved the `Must be a valid URL or data URI` key later in the file.

* **No reported changes**
  * Apps (standalone / OCP plugin), Cypress E2E tests, Go auth proxy, container/packaging, or CI/workflows were not part of this update based on the provided change set.


jkilzi (approved):
Clean, consistent removal. All related code paths (Yup validation, initial values, `getCatalogPatches`, `getCatalogResource`) are correctly updated.

Two minor notes:
- **Existing icon data preserved on edit** — because the patch is never generated, catalogs that already have an `icon` set will silently retain it. This is correct behaviour for a soft deprecation, but worth a comment if it's intentional.
- **`"Must be a valid URL or data URI"` i18n string** — it's moved rather than deleted. Confirm it is still referenced elsewhere; if not, it can be removed to keep the translation file clean.

PR #687, opened 2026-06-10
EDM-4143: Modified queued status icon

celdrake:
Table showing all possible statuses. 
Icon has been replaced with `PendingIcon` which looks like an hourglass. 

queued-icon

## Summary

**Area Affected:** `libs/ui-components/` (shared UI components)

**Changes:**
- Updated the image build and export status display for queued/pending states to render a dedicated PatternFly `PendingIcon` (hourglass) rather than the previous generic queued indicator.
- Added support for passing a per-status custom icon through the shared status rendering component.

**Technical Implementation (key file):**
- `libs/ui-components/src/components/ImageBuilds/ImageBuildAndExportStatus.tsx`
  - Introduces a `getQueuedStatusInfo(t)` helper that returns the queued status label/level along with `customIcon: PendingIcon`
  - Updates `getImageBuildStatusInfo()` and `getImageExportStatusInfo()` so pending/queued “reason” paths use `getQueuedStatusInfo()`
  - Extends the shared `ImageBuildAndExportStatus` renderer with an optional `customIcon` prop and forwards it into `StatusDisplayContent`
  - Updates `ImageBuildStatusDisplay` and `ImageExportStatusDisplay` to extract and pass through `customIcon` from their status-info helpers

**Cross-Cutting Impact:**
- This is a shared UI component change that should affect any screen(s) in the standalone app and the OCP plugin that reuse the image build/export status UI.

**Scope:**
- Localized to the shared UI component; no changes indicated for `libs/types/`, `libs/i18n/`, `libs/cypress/`, `apps/*`, the Go auth proxy, container builds, packaging, or CI/workflow configuration.

### DIFF
--- a/libs/ui-components/src/components/ImageBuilds/ImageBuildAndExportStatus.tsx
+++ b/libs/ui-components/src/components/ImageBuilds/ImageBuildAndExportStatus.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { Stack, StackItem } from '@patternfly/react-core';
 import { TFunction } from 'react-i18next';
+import { SVGIconProps } from '@patternfly/react-icons/dist/js/createIcon';
+import { PendingIcon } from '@patternfly/react-icons/dist/js/icons/pending-icon';
 
 import {
   ImageBuildCondition,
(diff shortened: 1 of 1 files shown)


jkilzi (approved):
Good change — `PendingIcon` is a much more descriptive visual than the generic unknown-level icon for a queued state.

Extracting `getQueuedStatusInfo` also removes the duplication between `getImageBuildStatusInfo` and `getImageExportStatusInfo`.

**Nit**: `customIcon` is typed as `React.ComponentClass<SVGIconProps>`. This is correct for current PF class-based icons but would exclude functional SVG components if PF ever migrates them. Low risk for now.

PR #690, opened 2026-06-10
EDM-4139: Remove shouldFocusFirstItemOnOpen as it does not react to hover

celdrake:
Removed `shouldFocusFirstItemOnOpen` since it always kept the focus on the first item while the menu was open, regardless of whether it was the selected item and/or if the cursor was over other items.

Made-with: Cursor

## Summary

This PR removes the `shouldFocusFirstItemOnOpen` focus-management prop from shared PatternFly `Select`/dropdown usages in `libs/ui-components/`, fixing the behavior where the first item stayed focused even when the user hovered other items.

## Areas Affected

**libs/ui-components/** (shared UI components):
- Updated `Select` focus handling in multiple components by removing `shouldFocusFirstItemOnOpen`, and using `shouldFocusToggleOnSelect` where applicable:
  - `libs/ui-components/src/components/DetailsPage/DetailsPageActions.tsx`
  - `libs/ui-components/src/components/Device/DeviceDetails/DeviceLogsLevelField.tsx`
  - `libs/ui-components/src/components/Device/DeviceDetails/DeviceLogsTimeRangeField.tsx` (also removed the explicit `shouldFocusToggleOnSelect={false}` override)
  - `libs/ui-components/src/components/form/FilterSelect.tsx`
  - `libs/ui-components/src/components/form/FormSelect.tsx` (replaces focus prop and adds explicit dropdown close on option select via `setIsOpen(false)`)

## Cross-Cutting Implications

Because these are shared UI components, the dropdown focus behavior change affects all screens using these `Select` implementations across the product experience (including both standalone and OCP-plugin usages).

### DIFF
--- a/libs/ui-components/src/components/DetailsPage/DetailsPageActions.tsx
+++ b/libs/ui-components/src/components/DetailsPage/DetailsPageActions.tsx
@@ -128,7 +128,6 @@ const DetailsPageActions = ({ children }: React.PropsWithChildren) => {
       onOpenChange={(isOpen: boolean) => setActionsOpen(isOpen)}
       popperProps={{ position: 'left', preventOverflow: true }}
       shouldFocusToggleOnSelect
-      shouldFocusFirstItemOnOpen
       toggle={(toggleRef) => (
         <MenuToggle
           ref={toggleRef}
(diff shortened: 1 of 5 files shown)


jkilzi (approved):
Correct removal. `shouldFocusFirstItemOnOpen` was conflicting with hover-based item highlighting, so dropping it from all Select usages is the right call.

**Note** — `DeviceLogsTimeRangeField` previously had both `shouldFocusToggleOnSelect={false}` and `shouldFocusFirstItemOnOpen={false}` explicitly set. After this PR both are removed and default to PF's default values. If PatternFly's default for `shouldFocusToggleOnSelect` is `true`, focus will return to the toggle after selection — confirm that is the desired behaviour for this particular field.

PR #692, opened 2026-06-11
EDM-4153: Improve layout of Repository select dropdown

celdrake:
The layout looks better now, with proper spacing between the selection mark and the additional content.

Removed `validateRepoSelection` as we had made changes to the RepositoryForm to enable only creating valid repositories (for example, only writable OCI registries).

repo-layout

Updated shared UI components in `libs/ui-components/` to improve the Repository select dropdown layout and simplify repository selection behavior.

- `libs/ui-components/src/components/form/FormSelect.css`: adjusts menu item spacing and positions the PatternFly select checkmark absolutely so it doesn’t overlap option content.
- `libs/ui-components/src/components/form/RepositorySelect.tsx`: removes `validateRepoSelection` support (and the related “invalid option”/error description generation). Repository creation/selection now directly commits the chosen/created repository value.
- `libs/ui-components/src/components/ImageBuilds/CreateImageBuildWizard/steps/OutputImageStep.tsx`: removes the step-level `writableRepoValidation` callback and relies on the updated helper text/RepositorySelect filtering behavior for OCI usage.

Cross-cutting impact:
- Because `RepositorySelect` is shared, this affects the image build wizard UI in both `apps/standalone/` and `apps/ocp-plugin/` (via shared `CreateImageBuildWizard` components).

- `libs/i18n/locales/en/translation.json`: updated text to align with the new “Only OCI-compliant registries are shown…” helper message.

No changes were made to the Go auth proxy, container/packaging, CI/workflows, or Cypress/E2E tests.

### DIFF
--- a/libs/i18n/locales/en/translation.json
+++ b/libs/i18n/locales/en/translation.json
@@ -1214,7 +1214,6 @@
   "Build image": "Build image",
   "You do not have permissions to promote builds to the catalog": "You do not have permissions to promote builds to the catalog",
   "Add to the software catalog testing channel upon successful build": "Add to the software catalog testing channel upon successful build",
-  "Repositories used for output images must be writable.": "Repositories used for output images must be writable.",
   "Management-ready by default": "Management-ready by default",
   "The agent is automatically included in this image. This ensures your devices are ready to be managed immediately after they are deployed.": "The agent is automatically included in this image. This ensures your devices are ready to be managed immediately after they are deployed.",
   "Target repository": "Target repository",
--- a/libs/ui-components/src/components/ImageBuilds/CreateImageBuildWizard/steps/OutputImageStep.tsx
+++ b/libs/ui-components/src/components/ImageBuilds/CreateImageBuildWizard/steps/OutputImageStep.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { Alert, Content, FormGroup, FormSection, Gallery } from '@patternfly/react-core';
 import { FormikErrors, useFormikContext } from 'formik';
 
-import { OciRepoSpec, RepoSpecType, Repository } from '@flightctl/types';
+import { RepoSpecType } from '@flightctl/types';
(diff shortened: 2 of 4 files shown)


jkilzi (approved):
Good layout overhaul. The `Grid` + `selectedLabel` pattern is a clear improvement over the old `Flex`-in-description approach.

**`validateRepoSelection` removal**: safe to drop — `writeAccessOnly: true` already filters the list so read-only repos never appear. The client-side double-check was redundant. The `handleCreateRepository` path is also clean since the create modal enforces write access.

**Nit — CSS comment**: the `.pf-v6-c-menu__item-select-icon` absolute-position override is correct but fragile against PatternFly version bumps. Add a short comment explaining the intent, e.g. `/* PF v6 checkmark overflows the Grid layout without this — position it absolutely so it doesn't affect column widths */`.

**Watch — merge ordering**: this PR, #698, and #702 all touch `RepositorySelect.tsx` and/or `ConfigWithRepositoryTemplateForm.tsx`. Recommended merge order: this PR first (biggest refactor), then #698, then #702 — each subsequent branch just needs a rebase to pick up the prior changes.

celdrake:
@jkilzi I rebased and fixed the conflicts with previous branches, PTAL when you can.

jkilzi (approved):
The `Grid` + `selectedLabel` pattern is the right call — the dropdown shows rich content (name, URL, accessibility status) while the trigger stays clean. Filtering via `writeAccessOnly: true` upfront is a UX improvement over the old validate-on-select approach.

The CSS checkmark fix is necessary and correct for the grid layout.

**Nit — dead code left behind.** `ReadOnlyRepositoryListItem` and the `invalidRepoItems` accumulator in `getRepositoryItems` are now unreachable: the `selectionError` branch was the only writer to `invalidRepoItems`, and it's been removed. The component still renders but always receives an empty map. Worth cleaning up in a follow-up.

celdrake:
I'll remove the dead code in the next PR, thanks for pointing that out.

PR #694, opened 2026-06-11
EDM-3535: Correct isRHEM detection for OCP plugin

celdrake:
The OCP plugin did not receive the "settings" therefore the "isRHEM" detection didn't work there.

## Areas affected

- apps/ocp-plugin/: Platform-specific OCP plugin frontend
- proxy/: Go auth proxy backend (UI settings endpoint + RHEM index selection)
- libs/ui-components/: Read-only impact (consumers depend on `settings.isRHEM`)

## Summary

This PR fixes `isRHEM` detection for the OCP plugin by exposing runtime UI settings from the Go proxy and consuming them in the plugin’s app context.

### apps/ocp-plugin/
- Updated `apps/ocp-plugin/src/components/AppContext/AppContext.tsx`:
  - Added a local `useUiSettings()` hook that initializes `isRHEM` from `window.isRHEM`, then fetches `${apiProxy}/ui-settings` (AbortController-protected).
  - Coerces `data.isRHEM` to a boolean (`!!data.isRHEM`), writes it back to `window.isRHEM`, and updates React state; defaults remain if the fetch fails/aborts.
  - `useValuesAppContext` now provides `settings: { isRHEM: boolean }` instead of a hardcoded empty `settings` object.
- Updated `apps/ocp-plugin/src/utils/apiCalls.ts`:
  - `uiProxy` is now exported so other modules can reuse the computed UI proxy base URL.

### proxy/
- Updated `proxy/config/config.go`:
  - `IsRHEM` now parses `IS_RHEM` as a boolean (`parseBoolEnv("IS_RHEM", false)`).
- Added `proxy/server/settings.go`:
  - Introduces `UISettingsHandler`, returning JSON `{ "isRHEM": <bool> }` sourced from `config.IsRHEM`.
- Updated `proxy/app.go`:
  - Conditionally registers `GET /api/ui-settings` only when `config.OcpPlugin == "true"`.
- Updated `proxy/server/server.go`:
  - `serveIndexPage` now checks `config.IsRHEM` as a boolean to choose between `index` and `index-rhem`.

## Cross-cutting implications

- The OCP plugin now relies on a new runtime proxy endpoint (`/api/ui-settings`) to determine `isRHEM`, aligning plugin behavior with server/env configuration.
- While this PR does not modify `libs/ui-components/`, it enables consistent `settings.isRHEM` availability to components that already consume it.
- No changes were made to `libs/types/`, `libs/i18n/`, `libs/cypress/`, `apps/standalone/`, packaging, container builds, or CI/E2E configuration.

### DIFF
--- a/apps/ocp-plugin/src/components/AppContext/AppContext.tsx
+++ b/apps/ocp-plugin/src/components/AppContext/AppContext.tsx
@@ -23,7 +23,7 @@ import { Prompt } from 'react-router-dom';
 import { getUser } from '@openshift-console/dynamic-plugin-sdk/lib/app/core/reducers';
 import { useSelector } from 'react-redux';
 import { useFetch } from '../../hooks/useFetch';
-import { uiProxy } from '../../utils/apiCalls';
+import { apiProxy } from '../../utils/apiCalls';
 
 import '@flightctl/ui-components/src/styles/global.css';
 import './AppContext.css';
@@ -34,7 +34,7 @@ const useUiSettings = (): { isRHEM: boolean } => {
   React.useEffect(() => {
     const controller = new AbortController();
 
-    fetch(`${uiProxy}/ui-settings.json`, { signal: controller.signal })
+    fetch(`${apiProxy}/ui-settings`, { signal: controller.signal })
       .then((response) => (response.ok ? response.json() : Promise.reject(new Error('Failed to load UI settings'))))
       .then((data: { isRHEM?: boolean }) => {
         const isRHEM = !!data.isRHEM;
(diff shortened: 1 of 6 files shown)


jkilzi (approved):
Solid fix. The two-part approach (Go bool + runtime `/ui-settings` fetch) is the right solution for OCP plugin RHEM detection.

**Go side**: converting `IS_RHEM` from `getEnvVar` (string) to `parseBoolEnv` (bool) eliminates the fragile `== "true"` check in `serveIndexPage`. Clean.

**Frontend**: `useUiSettings` correctly seeds from `window.isRHEM` as a fallback and uses `AbortController` for effect cleanup — no memory leak. Error is swallowed silently which is the right choice for a non-critical settings fetch.

**Note — auth**: `/ui-settings` returns `isRHEM` with no explicit auth check. This is fine since the entire OCP plugin proxy sits behind OpenShift auth anyway.

**Nit**: the initial `useState({ isRHEM: window.isRHEM ?? false })` means that if the RHEM index page already set `window.isRHEM = true`, there will be a harmless double render (initial true → fetch confirms true). Worth a comment for future readers.

PR #696, opened 2026-06-12
EDM-4152: Make all form alerts full width

celdrake:
Alerts within forms were not full-width, now we can use `FlightCtlFormAlert` for that purpose.
Before:

form-alerts-before

After:
form-alerts


celdrake:
We're clarifying exact requirements with UX.

celdrake:
Checked with UX and https://github.com/flightctl/flightctl-ui/pull/724 is a better solution.

PR #704, opened 2026-06-16
EDM-4145: Common helperText for unique Name field

celdrake:
Show a unified helperText in the "Name" field.
Add a clarification for CatalogItem's "Display Name".

## Areas Affected

**libs/ui-components/** (Shared UI Components)
- **Shared helper text behavior for “Name” fields** via the reusable form components:
  - `components/form/NameField.tsx`
  - `components/form/RichValidationTextField.tsx`
- **Catalog item “Display name” field UI clarification** in the Add Catalog Item wizard/modals:
  - `components/Catalog/AddCatalogItemWizard/CreateCatalogModal.tsx`
  - `components/Catalog/AddCatalogItemWizard/steps/GeneralInfoStep.tsx`
  - `components/Catalog/AddCatalogItemWizard/steps/TypeConfigStep.tsx`
- Minor **accessibility/props consistency** updates for “Display name” related fields:
  - `components/ImagePromotion/ImagePromotionForm.tsx`
- Removes duplicated/inline helper text from a consumer:
  - `components/AuthProvider/CreateAuthProvider/CreateAuthProviderForm.tsx`

**Cross-cutting Implications**
- Changes to **shared UI components** (`NameField`, `RichValidationTextField`) affect any app surfaces that consume `ui-components` (e.g., both standalone and any OCP plugin usage).

## Summary of Changes

This PR centralizes and unifies helper-text rendering for “Name” fields across the UI component library and clarifies “CatalogItem” “Display name” labeling.

**Core component updates**
- **`NameField`**
  - Adds support for a `helperText` prop and forwards it to `RichValidationTextField`.
  - Updates `NameFieldWrapper` so it always supplies the translated “A unique system identifier. Cannot be changed after creation.” helper text (replacing the prior conditional behavior where helper text was only rendered when provided/disabled).
  - Keeps the existing disabled/editability logic intact.
- **`RichValidationTextField`**
  - Adds an optional `helperText?: React.ReactNode` prop.
  - Renders the helper text via `DefaultHelperText` alongside existing input/validation UI.

**Consumer updates**
- **Auth provider form**: removes the redundant “can’t change after it’s created” helper text from the `NameField` usage so it relies on the centralized `NameField` behavior.
- **Catalog item wizard (“Display name”)**
  - Clarifies labeling for the “Display name” concept (including changing “Name”/accessibility text to “Display name” in the Type/Config step).
  - Removes the previous “Optional display name” helper text where applicable.
  - Reworks some JSX formatting and ensures relevant props (e.g., `aria-label`, `helperText`, `isDisabled`) are consistently applied.
- **Image promotion form**
  - Ensures “Display name” fields include appropriate `aria-label`, helper text, and explicit disabled handling; also adds an `aria-label` for the “Readme” field.

**Internationalization**
- The unified “Name” helper text is sourced from the i18n/translated message used by the shared helper rendering path (`DefaultHelperText`), ensuring consistent wording across the app.

## Non-goals / Explicitly out of scope
- No changes reported to `libs/types/`, `libs/i18n/` (beyond consuming existing translation wiring in the shared UI components), `libs/cypress/`, `apps/*`, Go auth proxy, container builds, packaging, or CI/workflow configuration.


libs/ui-components/src/components/form/NameField.tsx
```diff
@@ -69,21 +69,30 @@ const NameField = ({ name, validations, resourceType, ...rest }: NameFieldProps)
...
     );
   }
-  return <NameField name={name} validations={validations} resourceType={resourceType} {...rest} />;
+  return (
+    <NameField
+      name={name}
+      validations={validations}
+      resourceType={resourceType}
+      helperText={t('A unique system identifier. Cannot be changed after creation.')}
+      {...rest}
```
jkilzi, line 92: `{...rest}` is spread **after** the explicit `helperText` prop, so any caller that passes `helperText` through `rest` will override the standardised string — defeating the purpose of this PR.

Move `{...rest}` before `helperText` so the hardcoded text always wins:

```tsx
<NameField
  name={name}
  validations={validations}
  resourceType={resourceType}
  {...rest}
  helperText={t('A unique system identifier. Cannot be changed after creation.')}
/>
```

jkilzi (changes requested):
Good intent — standardising the  helper text is the right call. One prop-ordering bug needs fixing before merge.

**Accessibility wins** — the  additions in  and friends are a nice bonus.

**Nit** —  drops the *"Optional display name"* helper text from the artifact name field while also relabelling it "Display name". If the field is optional, some hint to that effect would still be helpful for users.

libs/ui-components/src/components/form/NameField.tsx
celdrake: You are right, fixed now.

jkilzi (approved):

PR #707, opened 2026-06-17
EDM-4150: Making placeholder styling more visually distinguishable

celdrake:
Adding new global style for placeholders, with reduced opacity.
Content changed in places where they were examples but without "e.g." or other similar text.

Applied the new style across the entire application, which required defining a common `FlightCtlModal` as PatternFly renders modals outside the app root HTML element.

## Summary

This PR improves placeholder styling across the UI (reduced placeholder opacity and more consistent “example” wording) and standardizes modal rendering by introducing a shared `FlightCtlModal` wrapper. This ensures the new globally scoped placeholder styles also apply to PatternFly portaled modals.

### libs/ui-components/ (Shared UI Components)
- **New `FlightCtlModal` wrapper**: Adds `FLIGHTCTL_APP_CLASS` (`fctl-app`) to the underlying PatternFly modal’s `className`, and re-exports `ModalProps` for typing.
- **Global placeholder styling**: Adds `libs/ui-components/src/styles/global.css` with `.fctl-app`-scoped rules to set placeholder opacity:
  - `.pf-v6-c-form-control > ::placeholder`
  - `.pf-v6-c-text-input-group__text-input::placeholder`
- **FilterSelect placeholder styling**:
  - Adds `.fctl-filter-select__placeholder { opacity: 0.6; }`
  - Updates `FilterSelect` to render the placeholder with that CSS class.
- **Placeholder copy/i18n updates**: Updates many `TextField`/`TextInput` placeholders to use translation-driven strings and example-style wording (e.g. `t('e.g. ...')` / `nsSeparator: '|'`) and a few accessibility-related label/placeholder tweaks.
- **Modal migrations (cross-cutting)**: Replaces direct PatternFly `Modal` usage with `FlightCtlModal` across many feature modals (auth, catalog, devices, images, fleets, repositories, resource sync, and various mass-action modals), so all of them inherit the same placeholder styling.

### apps/standalone/ (Platform-Specific App)
- **App-wide styling hookup**:
  - Imports shared `@flightctl/ui-components/src/styles/global.css`
  - Wraps `AppRouter` with a top-level `<div className={FLIGHTCTL_APP_CLASS}>` so the `fctl-app` scope exists in the DOM.
- **Logout modal**: Switches the logout failure UI to `FlightCtlModal`.

### apps/ocp-plugin/ (Platform-Specific App)
- **App-wide styling hookup**:
  - Imports shared `@flightctl/ui-components/src/styles/global.css` in `AppContext`
  - Updates `WithPageLayout` to render a `FLIGHTCTL_APP_CLASS` wrapper (`style={{ display: 'contents' }}`) around both the early-return and normal layout paths.

### eslint configuration
- Updates `eslint.config.js` `no-restricted-imports` to disallow direct imports of PatternFly `Modal`, directing callers to use `FlightCtlModal` instead.

## Cross-Cutting Implications
- **Shared UI change affects both platforms**: both standalone and OCP plugin ensure the `fctl-app` scope exists even with PatternFly’s portaled modals, by combining:
  - `FLIGHTCTL_APP_CLASS` wrappers at the app/layout level
  - the `FlightCtlModal` wrapper that appends the class to modal DOM.
- **Placeholder styling consistency**: placeholders inside both in-tree UI and portaled modals are now visually consistent (reduced opacity), and many placeholders were updated to explicitly reflect example-style wording (“e.g.”).

## Release
- Marked for inclusion in release **1.3**.


apps/ocp-plugin/src/components/common/WithPageLayout.tsx
```diff
@@ -27,7 +28,9 @@ const WithPageLayout = ({ children }: React.PropsWithChildren) => {
     <OrganizationGuard>
       <PermissionsContextProvider>
         <SystemRestoreProvider>
-          <WithPageLayoutContent>{children}</WithPageLayoutContent>
+          <div className={FLIGHTCTL_APP_CLASS}>
```
celdrake: Evaluating this extra div as it breaks the height for pages with "small" content, for example for Wizard steps with few fields.

celdrake: Fixed by using `display:contents` on it (via CSS)

libs/ui-components/src/components/Repository/CreateRepository/CreateRepositoryForm.tsx
```diff
@@ -272,7 +272,7 @@ const RepositoryType = ({ isEdit }: { isEdit?: boolean }) => {
         </SplitItem>
       </Split>
       {showConfirmChangeType && (
-        <Modal variant="small" isOpen>
+        <FlightCtlModal variant="small" isOpen>
           <ModalHeader title={'Change repository type?'} titleIconVariant="warning" />
```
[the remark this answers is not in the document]
celdrake: Done

libs/ui-components/src/styles/global.css
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/*
+ * Reduce opacity of all placeholders so they are clearly distinguishable from entered values.
+ */
+.fctl-app .pf-v6-c-form-control ::placeholder,
+.fctl-app .pf-v6-c-form-control textarea::placeholder,
+.fctl-app .pf-v6-c-text-input-group__text-input::placeholder {
+  opacity: 0.6;
+}
```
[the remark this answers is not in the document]
celdrake: Done

apps/ocp-plugin/src/components/common/WithPageLayout.tsx
celdrake: This was breaking the correct display in Modals. 
Now we inline the `display: contents` style only where it's needed.

jkilzi (approved):
Solid, well-scoped approach. The two-part strategy — scoping placeholder styles via `.fctl-app` + a `FlightCtlModal` wrapper for portaled modals — correctly handles the fact that PF modals render outside the app root.

**Good: ESLint enforcement.** The `no-restricted-imports` rule for `Modal` prevents regressions at import time — this is exactly how wrapper patterns should be enforced.

**Good: `FilterSelect` handled separately.** The `MenuToggle` placeholder isn't inside a `pf-v6-c-form-control`, so the global CSS wouldn't reach it. The dedicated `.fctl-filter-select__placeholder` class is the right local fix.

**Note — `display: contents` in OCP plugin.** `WithPageLayoutRoot` uses `display: contents` to avoid layout impact, which is correct. Worth confirming the target browser matrix covers it — `display: contents` has known historical issues with accessibility trees in older browsers.

**Note — standalone app DOM change.** `<AppRouter>` is now wrapped in a `<div className="fctl-app">`. Verify no existing CSS relies on `AppRouter` being the direct child of its provider, though this is low risk in practice.

**Nit — bonus i18n fix.** The hardcoded `title={'Change repository type?'}` in `RepositoryType` is now correctly wrapped in `t()`. Good catch — worth mentioning in the PR description so translators notice it.

PR #723, opened 2026-07-06
EDM-3804 Add confirmation screen

celdrake:
The bot in the previous PR added a commit after an intermitent network failure.

To stop adding more unnecessary costs, etc, I took the two valid commits to this new PR unrelated to the bot.

- **libs/ui-components/**: Updated `CreateCatalogModal` in the shared Catalog add wizard flow to show an in-modal success/confirmation state after a successful create instead of immediately closing. The modal now stores the created catalog in local state, switches to a “Catalog created” success view, and changes the footer to a single Close action.
- Removed the previous helper text from the **Display name** field in the create form.
- **Behavior notes**: Edit mode still returns success immediately on patch; read-only mode still blocks submission.

**Cross-cutting impact**
- This is a **shared UI component** change, so it affects both the standalone app and the OCP plugin wherever this catalog creation modal is reused.


celdrake:
Sorry @jkilzi would you mind re-approving from PR 700?
The bot is doing strange things and I don't want to make further changes in the original PR. 

I reported the issue to the bot creators. 

Thanks!

jkilzi (approved):

PR #725, opened 2026-07-07
Make validations for resource limits visible

celdrake:
After https://github.com/flightctl/flightctl-ui/pull/463, the Yup schema was incorrect and the errors where not appearing in the UI. Updated the message to be more helpful to the user.

Screenshot From 2026-07-07 16-50-46

Updated the shared UI validation schema in `libs/ui-components/` for container app resource limits.

- Changed resource limit validation from nested `limits.cpu` / `limits.memory` to new top-level `cpuLimit` / `memoryLimit` fields in the `validApplicationsSchema` Yup schema (keeping the existing CPU/memory validator functions, with updated schema shape and test names so errors display correctly).
- Updated the related shared types and form plumbing to match the new field names (e.g., `libs/ui-components/src/types/deviceSpec.ts`, and device spec utilities that map between legacy `limits.*` and the new `cpuLimit`/`memoryLimit` fields).
- Improved the user-facing validation messages (and ensured they’re backed by `libs/i18n/`), updating the displayed “CPU limit is invalid…” / “Memory limit is invalid…” text.

Cross-cutting impact: this is a shared UI component change used by both `apps/standalone` and `apps/ocp-plugin` flows via the shared device/fleet wizards. E2E/Cypress coverage and CI configuration were not indicated as directly impacted by this change.


jkilzi (approved):
Validated on workshop OCP (cluster-vzhx4) with a per-PR UI image build/deploy (`flightctl-ui-pr:pr-725`).

Invalid CPU/memory limit values surface as field errors (`aria-invalid` + helper text) and keep Next disabled. The Yup schema fix for flat `cpuLimit` / `memoryLimit` behaves correctly in the Create Fleet wizard.

PR #728, opened 2026-07-15
EDM-4782: Fix navigation for tabbed components in OCP v4.22

celdrake:
After the changes from OCP v4.22 related to the React Router, whenever the user would click on a new tab, the tab element was being concatenated to the previous URL, resulting in the tab content being empty.

The change should work in OCP v4.22 and also older versions.

## Summary

- Updated the shared `libs/ui-components/` `TabsNav` component to build **absolute** tab navigation paths compatible with React Router behavior changes introduced in **OCP v4.22** (and working with older versions).
- `TabsNav` now:
  - Derives the active tab from `location.pathname`.
  - On tab selection, trims any trailing slash and computes `nextPath` as an absolute path—**replacing the last URL segment** when already on an existing tab route, otherwise **appending** `/${tabKey}`—preventing URL concatenation that could lead to empty tab content.
- Cross-cutting impact: this shared component affects tabbed details pages in both **`apps/standalone/`** and **`apps/ocp-plugin/`**, because the OCP plugin wrappers render the same shared details pages/components that use `TabsNav`.
- No changes impact `libs/types/`, `libs/i18n/`, `libs/cypress/`, the Go/auth proxy (`proxy/`), container builds/packaging (`packaging/`), or CI configuration (`.github/workflows/`).

### DIFF
--- a/libs/ui-components/src/components/TabsNav/TabsNav.tsx
+++ b/libs/ui-components/src/components/TabsNav/TabsNav.tsx
@@ -25,18 +25,16 @@ const TabsNav: React.FC<TabsNavProps> = ({ children, 'aria-label': ariaLabel, ta
     setActiveTab(String(newActiveTab));
   }, [location.pathname, tabKeys]);
 
-  // Handle tab selection and navigate to the corresponding route
+  // Navigate to the selected tab path using absolute path from location.pathname
+  // Relative navigate() resolves differently based on the React Router version.
   const handleTabSelect = (_event: React.MouseEvent | React.KeyboardEvent | undefined, tabIndex: string | number) => {
     const tabKey = String(tabIndex);
     setActiveTab(tabKey);
 
-    const isOnTabPath = tabKeys.some((key) => location.pathname.endsWith(`/${key}`));
-    if (isOnTabPath) {
-      // React Router v6/v7 appends bare segments; navigate to sibling tab instead.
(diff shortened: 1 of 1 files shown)


jkilzi (changes requested):
## Validation (workshop OCP `cluster-vzhx4`)

Retested with a fresh per-PR image: checkout `pr-728` → OpenShift build → Helm deploy `flightctl-ui-pr:pr-728` → browser check.

### Standalone regression

On fleet Details (`/devicemanagement/fleets/tab-test-fleet/details`), clicking **YAML** navigates to:

`/devicemanagement/fleets/yaml/details`

instead of keeping the fleet id. The fleet id is dropped; the index redirect then appends `/details`, and the UI shows “Fleet not found” for id `yaml`.

### Cause

`navigate(\`../${tab}\`, { relative: "path" })` resolves against `.../<fleetId>/details` (no trailing slash), so `..` removes the fleet id segment.

### Ask

Please use a navigation approach that works for both standalone and the OCP console plugin (e.g. replace only the last path segment, or `./${tab}` with a trailing-slash-safe base) and re-verify standalone sibling tab switches.

celdrake:
@jkilzi Changed the new path calculation to be absolute. Tested in OCP v4.22 and v4.21, both for Standalone and OCP plugin apps.

jkilzi (approved):

PR #733, opened 2026-07-21
EDM-4726: Design update to clarify VM fields

celdrake:
- Applies latest changes to UX design to clarify how certain fields for VM applications work.
- Improve the serial console: show a "connected" message whenever the connection is established, highlighted in green (we would have had a Spinner for serial console of Windows VMs until now).

### Affected areas
- **`libs/ui-components/`**: Updated shared VM application form and console UI.
  - **`ApplicationVmForm.tsx`**: Reworked the configuration-mode selector from a `Flex` + `Popover` help into a `FormGroup`-scoped `ToggleGroup` with inline helper text; updated Cloud-init user data, Credentials, and Password helper guidance to clarify first-boot-only behavior, credential overwrites, and serial console login behavior.
  - **`Terminal/Terminal.tsx`** + **`Terminal/AppTerminal.tsx`**: Added a localized, optional **green “connected”** message to the terminal session (suppresses the initial waiting spinner when provided).
  - **`Table/TablePagination.tsx`**: Pagination now returns `null` when `itemCount === 0`.
- **`libs/i18n/`**: Updated **`libs/i18n/locales/en/translation.json`** with the new/adjusted UX strings (including the serial console “connected” message and updated VM field helper texts).
- **Apps / packaging dependencies**:
  - **`apps/standalone/package.json`**, **`apps/ocp-plugin/package.json`**, **`libs/ui-components/package.json`**, **root `package.json`**, and **`package-lock.json`**: Updated **`js-yaml`** to `4.3.0` (via direct deps and root `overrides`), affecting dependency resolution for all consumers.

### Impact
- Changes to **`libs/ui-components`** are shared, so they affect both **standalone** and **OCP plugin** UIs that render these components.
- No changes observed in Go auth proxy, CI/CD workflows, or E2E/cypress-related code as part of this update.


libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationVmForm.tsx
```diff
@@ -312,7 +307,7 @@ const ApplicationVmForm = ({ index, isReadOnly }: { index: number; isReadOnly?:
                     <span>{t('Password')}</span>
                     <DefaultHelperText
                       helperText={t(
-                        'A password may be needed to log in through the serial console. Without one, console access may not be available.',
+                        'Sets a login password via cloud-init, required for serial console access. If not configured, the default credentials from the OS image apply (if any).',
```
asafbennatan: **Copy clarity:** This helper says the password is "required for serial console access," then that OS image defaults apply if it is not configured. Those statements conflict and undercut the clarify-fields goal.

Suggestion: soften to something like: "Sets a login password via cloud-init for serial console login. If not configured, the default credentials from the OS image apply (if any)." Update the `t()` literal and regenerate i18n.

libs/ui-components/src/components/Terminal/Terminal.tsx
```diff
@@ -146,13 +165,17 @@ const Terminal = React.forwardRef<ImperativeTerminalType, TerminalProps>(({ onDa
       pendingWriteRef.current = '';
       flushScheduledRef.current = false;
       terminal.current?.reset();
-      setReceivedData(false);
+      if (connectedMessageRef.current && terminal.current) {
+        writeConnectedMessage(terminal.current);
```
asafbennatan: **Reconnect timing:** `AppTerminal.handleReconnect` calls `reset()` while the connection is still closed (`isClosed` alert still visible). Writing the green "Connected…" banner here is premature; Terminal then unmounts when `isConnecting` flips, and the mount `useEffect` already writes the message after the WebSocket opens.

Suggestion: in `reset()`, when `connectedMessage` is set, clear the xterm buffer and keep the spinner suppressed (`setReceivedData(true)`) without calling `writeConnectedMessage`.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationVmForm.tsx
```diff
@@ -170,37 +167,31 @@ const ApplicationVmForm = ({ index, isReadOnly }: { index: number; isReadOnly?:
...
-              isDisabled={isReadOnly}
-              onChange={() => handleModeChange('yaml')}
-            />
-          </ToggleGroup>
-        </FlexItem>
-        <FlexItem>
-          <Popover
-            bodyContent={t(
-              'Use Form for standard Linux VMs with a container disk image, CPU, memory, and cloud-init. Use YAML for advanced configurations such as Windows VMs, custom networking, or other KubeVirt options not available in the form.',
+      <FormGroup fieldId={`${appFieldName}-config-mode`}>
```
asafbennatan: **A11y / consistency:** This `FormGroup` has a `fieldId` but no visible `label`. The ToggleGroup only has `aria-label={t('Configuration mode')}`.

Suggestion: add `label={t('Configuration mode')}` (key already exists), or drop `fieldId` if the group intentionally stays label-less.

asafbennatan (commented):
see below

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationVmForm.tsx
celdrake: removed the contradictory part.

libs/ui-components/src/components/Device/EditDeviceWizard/steps/ApplicationVmForm.tsx
celdrake: According to the design, there's no label. Removed the `fieldId`

libs/ui-components/src/components/Terminal/Terminal.tsx
celdrake: I couldn't see the problem you're describing, but I think it makes sense that we could skip writing the message on `reset`

libs/ui-components/src/components/Terminal/Terminal.tsx
```diff
@@ -31,14 +36,26 @@ export type ImperativeTerminalType = {
...
+
+  const writeConnectedMessage = React.useCallback((term: XTerminal) => {
+    const message = connectedMessageRef.current;
+    if (!message) {
+      return;
+    }
+    // Show connection message in green; reset so guest output is unaffected.
+    term.write(`\x1b[32m${message}\x1b[0m\r\n\r\n`);
+    setReceivedData(true);
+  }, []);
```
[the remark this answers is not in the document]
celdrake: `connectedMessage` is persistent for a Terminal instance, no need to handle this.

asafbennatan (approved):
LGTM

PR #748, opened 2026-07-22
EDM-4859: Better align the Catalog content to the page

celdrake:
Make the content for Catalog consistent with the rest of the pages. The outer wrapping element does not use full-height unless the content requires so.

See comparison with Image Builds page, and how the Catalog would look like for fewer or more catalog items.

With fewer items:
with-few

With many items:
with-many

- Updated the shared Catalog UI in `libs/ui-components/` to use the default `PageSection` layout instead of the wizard layout, improving content height and page alignment.
- Catalog rendering, filters, empty states, and split-pane behavior remain unchanged.
- No changes to `libs/types/`, `libs/i18n/`, `libs/cypress/`, platform-specific apps, auth proxy, packaging, or CI.
- As a shared component change, the updated layout may affect both standalone and OCP plugin consumers.

### DIFF
--- a/libs/ui-components/src/components/Catalog/CatalogPage.tsx
+++ b/libs/ui-components/src/components/Catalog/CatalogPage.tsx
@@ -234,7 +234,7 @@ export const CatalogPageContent = ({
               </div>
             </PageSection>
           ) : (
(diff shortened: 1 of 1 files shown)


jkilzi (approved):
LGTM

PR #664, opened 2026-06-02
EDM-3868: Reset page to 1 when filters change

celdrake:
Fixes filters not returning the expected results when the user was on a page other than page 1.

 Fixed for:
-  Main devices page
 - Decommissioned devices page
 - Catalog page for deploying an element to fleetless devices (simplified signature of `useDevicesPaginated` to make the code simpler while keeping all existing functionality)

## Summary by CodeRabbit

* **Bug Fixes**
  * Pagination resets to page 1 when device filters change.
  * The fix applies to the main devices page, the decommissioned devices page, and the catalog deployment flow.
  * Invalid filters do not trigger pagination changes.

* **Shared UI Components**
  * Added `useResetPaginationOnFilterChange` for reusable pagination state handling.
  * Simplified `useDevicesPaginated` to accept device name and package-mode filters.
  * Updated catalog device selection to use the revised pagination API.

* **Cross-Cutting Impact**
  * Changes are limited to `libs/ui-components/`.
  * Shared UI changes can affect both standalone and OCP plugin consumers.
  * No changes affect `libs/types/`, `libs/i18n/`, `libs/cypress/`, platform-specific app code, the Go auth proxy, container builds, E2E tests, or CI configuration.
  * No security impact is indicated. The change improves pagination correctness and maintainability.

PR #770, opened 2026-08-19
EDM-5267: Mitigate errors when UI service restarts

celdrake:
Ocassionally, the UI could show a "failed to load chunk" error instead of the selected page.
This seemed to happen when the UI service is restarted, as the loaded UI does not have the correct bundled assets.

Fixes:
- Whenever the route changes, re-import the component for the current page.
- If the "failed to load chunk" happens for any reason, we show a "Reload page" button which should fix the problem.
- The ErrorBoundary uses `location.key` so that when navigating to another tab after the UI has fully restarted (or back to the same tab), the page is shown instead of the stale error.

## Affected areas

- `apps/standalone/`: Re-imports route components when the route changes. Keys `ErrorBoundary` with `location.key` to clear stale errors after navigation or a UI service restart.
- `libs/ui-components/`: Adds `lazyResetter` and a “Reload page” action for chunk-loading errors.

## Impact

- Shared UI component behavior changes for consumers, including standalone and OCP plugin integrations.
- Standalone users can recover from outdated bundled assets without manual browser actions.
- No changes affect `libs/types/`, `libs/i18n/`, `libs/cypress/`, `apps/ocp-plugin/`, `proxy/`, `packaging/`, or CI configuration.
- No authentication or other security behavior changes are included.
- The change improves correctness by preventing stale lazy-loaded components and error boundaries from persisting across navigation.

### DIFF
--- a/apps/standalone/src/app/routes.tsx
+++ b/apps/standalone/src/app/routes.tsx
@@ -15,6 +15,7 @@ import {
   type RouteObject,
   RouterProvider,
   createBrowserRouter,
+  useLocation,
   useParams,
   useRouteError,
 } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


asmasarw (commented):
### 🔴 `resettableLazy.tsx` re-creates the lazy component whenever `location.key` changes:

In react-router v6, location.key is regenerated on every history transition (push and replace), not only when moving between pages.

Consequences confirmed:

1. Detail pages remount on every tab switch. DeviceDetailsPage (and Fleet/Repository/ImageBuild/AuthProvider details) render internal <Routes> under :id/*. Switching tabs changes location.key → the whole page remounts (refetch + lost tab state + spinner).
2. List pages remount when a URL-synced filter is applied. DevicesPage commits filters via setSearchParams (push) → new location.key → remount: search input loses focus mid-typing, pagination resets, full refetch. This also undercuts the pagination work in #664.

(Navigations run in a React transition, so the spinner may not always visibly flash — but the remount, state loss, and refetch still happen.)

### Suggested fixes (any one)

- **Simplest: drop the per-navigation re-import and rely on the ErrorBoundary "Reload page" button for recovery — a full reload already pulls fresh assets.**

- Retry-on-error: wrap the dynamic import() to catch a chunk-load failure and re-import (optionally cache-busting), so a fresh lazy is created only when a load actually fails:

```tsx
const load = () => importer().catch(() => importer());
const Component = React.useMemo(() => React.lazy(load), []);
```

- If reset-on-nav is truly desired, key it on the matched route/module (something stable per page), not location.key.

PR #779, opened 2026-08-31
Clarify further translation & styling rules

celdrake:
- After the last wording changes to i18n rules, Coderabbit was still adamant that a string such as `{{numVar1}} / {{numVar2}}` should be localized. 

The previous wording was more ambiguous, as it indicated that the copy needed to be "meaningfully localized by a translator". Using the "must have at least one word" leaves out the ambiguity. `{{numVar1}} of {{numVar2}}` should be localized, but the first example shouldn't.

- Also, clarify that using CSS such as  `var(--some-patternfly-token, #0FF)` is OK, as way to define a fallback when the PatternFly token may be unset.

## Affected area

- `.coderabbit.yaml` only.
- No changes affect `libs/`, `apps/`, `proxy/`, `packaging/`, shared UI components, platform-specific app code, the Go auth proxy, container builds, or E2E tests.
- This is a CI and code-review configuration change.

## Changes

- Classifies strings with literal words as translatable.
- Excludes punctuation-only and interpolation-only strings, such as `{{numVar1}} / {{numVar2}}`.
- Includes strings that combine words with variables, such as `{{numVar1}} of {{numVar2}}`.
- Permits CSS fallbacks for unset PatternFly tokens, such as `var(--some-patternfly-token, `#0FF`)`.
- Applies the same literal-word rules to the i18n compliance pre-merge check.

## Impact

- Review guidance and i18n checks now use consistent translation rules.
- CSS guidance supports valid PatternFly token fallbacks.
- The change does not alter runtime behavior or introduce cross-cutting application changes.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -157,10 +157,17 @@ reviews:
     - path: "{libs,apps}/*/src/**/*.{ts,tsx}"
       instructions: |
         i18n review:
-        - User-facing copy that a translator can meaningfully localize
-          must use the t() translation function. Translation changes
-          visual wording only — it must not alter UX behavior, layout
-          structure, or functional semantics.
+        - User-facing strings that contain at least one literal word
+          (alphabetic text outside {{interpolation}} placeholders) must
+          use the t() translation function. Translation changes visual
+          wording only — it must not alter UX behavior, layout structure,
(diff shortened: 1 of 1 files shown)


asmasarw (approved):
LGTM

PR #788, opened 2026-09-02
EDM-3334: Prevent pending device list flicker

eldar101:
## Summary
- Keep the pending enrollment requests section mounted while an unfiltered reload is in progress.
- Include enrollment request update/loading state so clearing a no-result search shows loading instead of briefly exposing the enrolled-devices empty state.
- Add stable Cypress test IDs for the pending enrollment search, section, and loading state, with coverage for clearing a no-result search.

## Notes
- Cypress validation used the standalone app with mocked API responses on titan52. I did not perform a live OCP/ACM console verification against a deployed plugin build.

## Summary

- **Shared UI:** Updated `libs/ui-components/` so the pending enrollment requests section remains mounted during an unfiltered reload. The section now shows its loading state instead of the enrolled-devices empty state.
- **E2E tests:** Updated `libs/cypress/` page objects, fixtures, and interceptors. Added Cypress coverage for clearing a pending enrollment request search with no results.
- **Platform applications:** No changes to `apps/standalone/` or `apps/ocp-plugin/`. The shared UI change may affect both applications.
- **Other areas:** No changes to `libs/types/`, `libs/i18n/`, `proxy/`, `packaging/`, or `.github/workflows/`.
- **Validation:** Cypress validation used a standalone app with mocked API responses on titan52. Live OCP/ACM console verification was not performed.

### DIFF
--- a/libs/cypress/e2e/devices/approveEnrollmentRequest.cy.ts
+++ b/libs/cypress/e2e/devices/approveEnrollmentRequest.cy.ts
@@ -3,6 +3,7 @@ import { ApproveEnrollmentRequestModalPage } from '../../pages/ApproveEnrollment
 
 let devicesPage: DevicesPage;
 let approveERModalPage: ApproveEnrollmentRequestModalPage;
+const FIRST_PENDING_ENROLLMENT_REQUEST_NAME = 'a021622d8633782719874da4052f957faa742fc7050026748bc79065c8819d139';
 
 describe('Enrollment requests approval', () => {
   beforeEach(() => {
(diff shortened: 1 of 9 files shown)


celdrake:
@eldar101 Thanks for the contribution!
Unfortunately, it seems to break the fix we recently did for EDM-3868. With the branch, if we're in a page other than the first and add some filters, we should be resetting to page 1 but that doesn't happen.

I reproduced it by adding a few fake devices and updating `PAGE_SIZE` to a low number, in case that helps.

eldar101:
> @eldar101 Thanks for the contribution! Unfortunately, it seems to break the fix we recently did for EDM-3868. With the branch, if we're in a page other than the first and add some filters, we should be resetting to page 1 but that doesn't happen.
> 
> I reproduced it by adding a few fake devices and updating `PAGE_SIZE` to a low number, in case that helps.

Thanks for the reply. I reproduced the pagination regression. I’ll update the hook to reset pagination when the search changes, while preserving the loading-state fix for clearing an empty search, and add coverage for both cases.
Let me know if that is ok or should I close the PR?

celdrake:
@eldar101 I think it's OK to keep this PR. Maybe it's worth first rebasing in case that changes anything?

eldar101:
@celdrake 

> @eldar101 I think it's OK to keep this PR. Maybe it's worth first rebasing in case that changes anything?

rebased and fixes. passes all checks.

PR #807, opened 2026-09-09
EDM-5888 display vulnerability backend

celdrake:
- Updated types to latest API changes
- Display the `source` for vulnerabilities, both in Fleet/Device details page, as well as in Overview.

vuln-v2

## Summary

- Updated shared vulnerability UI components to display scanner sources and improve grouped-vulnerability layouts.
- Added source mapping for Trustify, Quay, missing, and unknown sources.
- Prevented vulnerability severity details from wrapping.
- Exported `VulnerabilitySource` and delta-generation types from shared type indexes.
- No changes are reported for applications, i18n, Cypress, proxy, packaging, or CI.
- Security impact: improves scanner visibility in vulnerability details. It does not change vulnerability detection or remediation.
- Test results are unavailable.

## Risk classification

The applied risk label and labeling criteria are unavailable. The available evidence supports a shared UI and type-export change only.


libs/ui-components/src/utils/vulnerabilities.ts
```diff
@@ -51,6 +52,25 @@ export const getPrimaryVulnerabilityGroupFinding = (
...
+
+export const getVulnerabilitySource = (vulnerability: Vulnerability | VulnerabilityGroupItem, isRHEM: boolean) => {
+  switch (vulnerability.source) {
+    case VulnerabilitySource.Trustify:
+      return isRHEM ? DEFAULT_VULNERABILITY_SOURCE_DS : DEFAULT_VULNERABILITY_SOURCE_US;
+    case VulnerabilitySource.Quay:
+      return QUAY_VULNERABILITY_SOURCE;
+    case undefined:
+      // Source is missing, fallback to the default backend.
+      return DEFAULT_VULNERABILITY_SOURCE_US;
```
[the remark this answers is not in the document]
celdrake: When `source` is missing, the system will always return `Trustify` as that's what the system is currently always displaying. There should be a migration that should populate the data, but if it's missing for any reason, we'll keep the old behaviour.

EfratIfergan: Could we make the missing-source fallback respect the RHEM mode as well? When source is undefined (for example, with older or partial API responses), this branch always returns "Trustify", while the explicit VulnerabilitySource.Trustify branch maps to "Red Hat Trusted Profile Analyzer" in RHEM. Please use isRHEM ? DEFAULT_VULNERABILITY_SOURCE_DS : DEFAULT_VULNERABILITY_SOURCE_US here so legacy responses display the same product name.

celdrake: Changed. Rather than keeping the old behaviour, we'll always show TPA for Downstream and Trustify for Upstream.

EfratIfergan (approved):
LGTM

PR #808, opened 2026-09-10
EDM-5042: Small design adjustments

celdrake:
Follow-up for Device details redesign:

- Show "My field"  for `customInfo` fields, as it was being done before, and we still do for `systemInfo` fields
- Show "No value reported" for `customInfo` fields that don't have a value, as those are not skipped.
- Change `Label` color to blue when it refers to Device/Fleet labels selector
- Changed wording for application issues so it is valid for errors, warnings, and a mixture of both.

Screenshot From 2026-09-10 09-50-07

## Summary

- **Shared UI components (`libs/ui-components/`):**
  - Show `My field` for custom fields.
  - Show `No reported value` when a custom field has no value.
  - Use blue labels for Device and Fleet label selectors.
  - Use “application issues” wording for errors, warnings, or both.
  - Export `propNameToTitle` for shared field-name formatting.

- **Cross-cutting impact:**
  - The shared component changes affect all consumers, including the standalone application and OCP plugin.
  - No changes affect types, internationalization, proxy authentication, packaging, container builds, E2E tests, or CI configuration.
  - No security or data-flow behavior changes are introduced.

## Risk classification

**risk:ship** — The changes are limited to shared UI presentation, wording, and a helper export. They do not change APIs, persistence, authentication, authorization, or runtime data handling. The PR does not meet `risk:show` criteria because it does not introduce a broader behavioral or operational impact.


libs/ui-components/src/components/Device/DeviceDetails/DeviceCustomDataCard.tsx
```diff
@@ -21,12 +23,17 @@ const DeviceCustomDataCard = ({ customInfo }: { customInfo: [string, string][] }
...
-              <DescriptionListDescription>{entry[1]}</DescriptionListDescription>
-            </DescriptionListGroup>
-          ))}
+          {customInfo.map((entry, index) => {
+            const value = entry[1];
+            return (
+              <DescriptionListGroup key={index}>
+                <DescriptionListTerm>{propNameToTitle(entry[0])}</DescriptionListTerm>
+                <DescriptionListDescription>
+                  {value || <Content component="small">{t('No value reported')}</Content>}
```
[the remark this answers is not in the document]
celdrake: Updated the PR description, the designer's suggested copy is "No value reported"

libs/ui-components/src/components/Repository/RepositoryDetails/RepositorySourceList.tsx
```diff
@@ -52,7 +77,11 @@ export const RepositorySourcePlainList = ({ configs }: { configs: ConfigSourcePr
 
         return (
           <StackItem key={config.name}>
-            <RepositorySource config={config} repoDetails={repoDetails} />
+            {isRepoConfig(config) ? (
+              <RepositoryConfig config={config} repoDetails={repoDetails} showConfigName />
+            ) : (
+              <>{config.name}</>
```
liatb-rh, line 83: RepositorySourcePlainList  go over any non-repo config, so the same device will show e.g. "my-inline-config" in the plain list but "3 files" in the description list. 
The simplest fix is to replace the isRepoConfig ternary in RepositorySourcePlainList with <RepositorySourceItem config={config} repoDetails={repoDetails} />.

celdrake: Thanks for the review! AFAIK, this is according to the current design.

Device details --> This was redesigned to a two-column layout. Config names are displayed in the left column, and additional details on the right column. For inline configs, UX decided to show the number of files.


Fleet details --> This page wasn't redesigned, it has a single column for Configurations. Given that the config name quite be quite long, we can't also display the number of files.

I'm assuming at some point Fleet details page will be redesigned, in that case we could potentially display the same kind of details in both places.

inline-configs

